PR #8, opened 2019-07-16
Display as system tray again. Now with attachment to menu bar

JamieRuderman:


backend/src/Application.ts
```diff
@@ -203,6 +203,31 @@ class ElectronApp extends EventEmitter {
       }
     })
   }
+
+  showWindow() {
+    if (this.window && this.tray) {
```
danawoodman: What does this do exactly?

frontend/src/components/App/App.tsx
```diff
@@ -76,9 +76,9 @@ export const App = connect(
           className="w-100 h-100 df ai-stretch"
           style={{ flexFlow: 'column' }}
         >
-          <div className="df ai-center jc-center py-xs center dragable primary txt-md bg-gray-lighter">
+          {/* <div className="df ai-center jc-center py-xs center dragable primary txt-md bg-gray-lighter">
```
danawoodman: You removed logo/branding?

frontend/src/components/CopyableText/CopyableText.tsx
```diff
@@ -11,14 +11,14 @@ export function CopyableText({
   ...props
 }: CopyableTextProps & React.HTMLProps<HTMLDivElement>) {
   return (
-    <div {...props}>
+    <>
```
danawoodman: Why remove this? This allows you to style the component etc

backend/src/Application.ts
JamieRuderman: It just shows the window in the correct location. It needs to check for the presence of this.window and this.tray to make typescript happy. If those vars were initialized a little different it says that the wouldn't have to be optional params.

frontend/src/components/App/App.tsx
JamieRuderman: I just removed the title bar.

frontend/src/components/CopyableText/CopyableText.tsx
JamieRuderman: Because having the two components nested under a div was making them wrap poorly, this way all the elements are on the same level and can be handled consistently by the flexbox.

danawoodman:
This PR was not approved. Please make sure PR is approved before merging in the future. These changes have broken the build for me so I am reverting them

PR #10, opened 2019-08-06
[DESK-159] window placement

JamieRuderman:


frontend/src/pages/Page/Page.tsx
```diff
@@ -39,36 +40,52 @@ export interface Props {
...
-          left: 0,
-          right: 0,
-          overflow: 'hidden',
-        }}
-      >
-        {children}
-      </div>
+      {isMac() && (
+        <div
+          style={{
```
danawoodman: Why not move this to a variable like the above?

frontend/src/services/Platform.ts
```diff
@@ -10,3 +10,8 @@ export function isElectron() {
   const userAgent = navigator.userAgent.toLowerCase()
   return userAgent.includes(' electron/')
 }
+
+export function isMac() {
```
danawoodman: Don't you mean to just use `Environment` class like above?

frontend/src/pages/Page/Page.tsx
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
+import { isMac } from '../../services/Platform'
```
danawoodman: Maybe use `Environment` instead?

frontend/src/pages/Page/Page.tsx
JamieRuderman: Ok, was worried it wouldn't have the info when it loaded, but it seems to work fine.

frontend/src/services/Platform.ts
JamieRuderman: The environment is different... (electron vs browser) this could be considered the platform (mac, windows, linux) maybe. I don't want to try and detect all those right now since we don't need it yet. If we do we can update.

frontend/src/pages/Page/Page.tsx
JamieRuderman: Maybe you mean environment from the backend? I don't have that available in the front end.

PR #24, opened 2020-01-14
2.4.7 merge

JamieRuderman:
Keep user signed in on desktop but not from browser
Fix uninstalling bin directory on Mac/Linux


frontend/src/store.tsx
```diff
@@ -7,7 +7,7 @@ import logger from 'redux-logger'
 const plugins = [immerPlugin(), selectPlugin()]
 let middlewares: any[] = []
 
-if (process.env.NODE_ENV !== 'test') middlewares.push(logger)
+// if (process.env.NODE_ENV !== 'test') middlewares.push(logger)
```
danawoodman: Intentional?

frontend/src/models/auth.ts
```diff
@@ -72,7 +96,10 @@ export default createModel({
     async signedIn() {
       dispatch.auth.signInFinished()
       dispatch.auth.checkSignInFinished()
-      dispatch.devices.shouldSearchDevices()
+      if (!r3.token) await dispatch.auth.fetchAuthToken()
+      const searchOnly = await dispatch.devices.shouldSearchDevices()
+      console.log('SEARCH ONLY?', searchOnly)
```
danawoodman: Remove log

frontend/src/models/devices.ts
```diff
@@ -44,16 +44,13 @@ export default createModel({
...
-        dispatch.devices.fetch()
-        return
-      }
+      let searchOnly = pref !== 'false'
 
       // If they have too many services, show search only.
-      return r3.devices
-        .count()
-        .then(count => dispatch.devices.setSearchOnly(count.services > SEARCH_ONLY_SERVICE_LIMIT))
+      await r3.devices.count().then(count => (searchOnly = count.services > SEARCH_ONLY_SERVICE_LIMIT))
```
danawoodman: Why not use await, assign count to var and then do your variable setting? Seems more idiomatic async/await

frontend/src/store.tsx
JamieRuderman: No didn't mean to commit - backed it out already. Thanks!

frontend/src/models/devices.ts
JamieRuderman: Fixed!

PR #27, opened 2020-01-15
DESK-291 Check and restart auto-start connections every 30s

JamieRuderman:


backend/src/ConnectionPool.ts
```diff
@@ -36,6 +36,14 @@ export default class ConnectionPool {
     EventBus.on(ElectronApp.EVENTS.ready, this.updated)
   }
 
+  // maintain auto start connections
+  check = () => {
+    this.toJSON().map(c => {
+      // auto start if not running and has been started before
+      if (c.autoStart && !c.pid && c.startTime) this.start(c)
```
bstrech: wouldn't this autoStart happen everytime check is run since it is at the start of a boolean evaluation?

JamieRuderman: It will auto start if
1. The autoStart setting is checked
2. The connection isn't already running (no process id)
3. The startTime is undefined (start time is set the first time it is started)
I *think* the logic is right?

PR #31, opened 2020-01-22
DESK-238

JamieRuderman:
Was trying to re-connect to maintain online state when a service was offline. So now added online state to connection so that we don't try to connect if offline. Also allowed to update state in Tray menu and react appropriately.

### DIFF
--- a/backend/src/ElectronApp.ts
+++ b/backend/src/ElectronApp.ts
@@ -125,7 +125,6 @@ export default class ElectronApp {
   }
 
   private openWindow = (openDevTools?: boolean) => {
-    console.log('open window')
     if (!this.window || !this.tray) return
     d('Showing window')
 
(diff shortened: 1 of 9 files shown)


backend/src/ElectronApp.ts
```diff
@@ -125,7 +125,6 @@ export default class ElectronApp {
   }
 
   private openWindow = (openDevTools?: boolean) => {
-    console.log('open window')
```
bstrech: Do you really want to keep this console.log

backend/src/TrayMenu.ts
```diff
@@ -85,15 +90,19 @@ export default class TrayMenu {
         const location = hostName(connection, this.privateIP)
         result.push({
           label: connection.name,
-          icon: connection.active ? iconConnected : iconOnline,
+          icon: connection.active ? iconConnected : connection.online ? iconOnline : iconOffline,
```
bstrech: Ternary in a ternary.  Is that a pattern we want to keep?

JamieRuderman: I'm okay with it.  :)

backend/src/ElectronApp.ts
JamieRuderman: No - it's been removed in another pr: https://github.com/remoteit/desktop/compare/DESK-315

PR #33, opened 2020-01-24
DESK-308 Reduce heartbeat logging and provide a way to exec to CLI silently

JamieRuderman:


backend/src/CLI.ts
```diff
@@ -146,14 +146,25 @@ export default class CLI {
 
   async isNotInstalled() {
     const installed = RemoteitInstaller.fileExists(RemoteitInstaller.binaryName)
-    Logger.info('CLI INSTALLED?', { installed, name: RemoteitInstaller.binaryName })
+    d('CLI INSTALLED?', { installed, name: RemoteitInstaller.binaryName })
```
bstrech: is d a shortcut for debugger?

JamieRuderman: Yes, it is a debugger method that is conditional based on how you start the app.

PR #39, opened 2020-01-28
Desk 204 Building both deb and AppImage for linux

JamieRuderman:


frontend/package.json
```diff
@@ -10,9 +10,7 @@
     "eject": "react-scripts eject",
     "build-storybook": "build-storybook -s public",
     "storybook": "start-storybook -p 9009 -s public",
-    "type-css": "tcm -s -c -p src/**/*.module.css",
-    "watch-react": "react-scripts start",
-    "watch-type-css": "onchange -i -k 'src/**/*.module.css' -- npm run type-css"
```
bstrech: You dont want to watch for css changes?

JamieRuderman: Oh right, no I'm not using module.css files in desktop. I'm using the Material-UI framework for styling. So this can go.

PR #43, opened 2020-02-13
DESK-328 Separate electron out of backend and detect elevated permissions so it can run on a headless system

JamieRuderman:


backend/src/BinaryInstaller.ts
```diff
@@ -66,28 +67,34 @@ class BinaryInstaller {
...
   }
+
+  removeFile(commands: Command, path: string) {
+    if (!existsSync(path)) return
+    commands.push(environment.isWindows ? `del /Q /F "${path}"` : `rm -f ${path}`)
+  }
+
+  removeDir(commands: Command, path: string) {
+    if (!existsSync(path)) return
+    commands.push(environment.isWindows ? `rmdir /Q /S "${path}"` : `rm -rf ${path}`)
```
danawoodman: `rm -rf` is **VERRRRRY** dangerous, I would be pretty cautious with such code. Consider using a library instead like rimraf or similar.

JamieRuderman: Yea I actually did a full implementation with an npm module called trash, that would just move stuff to the trash but I couldn't get it to work. I would be happy to change this to a rimraf module, there might be an issue with running it with the sudo-prompt though. I'll poke around.

PR #53, opened 2020-02-22
DESK-321 Report device name in system info response

JamieRuderman:


backend/src/systemInfo.ts
```diff
@@ -0,0 +1,18 @@
+import environment from './environment'
+import cli from './cliInterface'
+import lan from './LAN'
+import os from 'os'
+
+export default async function systemInfo() {
+  await lan.getInterfaces()
+
+  let info: ILookup = {
```
danawoodman: Probably just return here instead of allocating a variable?

JamieRuderman: Yep! missed that one thanks!

PR #54, opened 2020-02-26
DESK-368 Adding support for Pi Zero armv6

JamieRuderman:


JamieRuderman:
Yea. They were so small I just snuck them in... change of border radius and margin.

PR #56, opened 2020-02-27
DESK-370 Clearer airbrake messages

JamieRuderman:
So we can attempt to filter out the heartbeat errors.

### DIFF
--- a/backend/src/Command.ts
+++ b/backend/src/Command.ts
@@ -64,7 +64,7 @@ export default class Command {
     } catch (error) {
       AirBrake.notify({
         params: { type: 'COMMAND ERROR', command: this.toString() },
-        error: error,
+        error,
       })
       this.log(`EXEC ERROR CAUGHT ${this.toString()}`, { error, errorMessage: error.message }, 'error')
       this.onError(error)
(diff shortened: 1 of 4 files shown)


backend/src/Command.ts
```diff
@@ -59,7 +62,10 @@ export default class Command {
         result = stdout.toString()
       }
     } catch (error) {
-      AirBrake.notify({ message: 'COMMAND ERROR', command: this.toString(), error })
+      AirBrake.notify({
+        params: { type: 'COMMAND ERROR', command: this.toString() },
+        error: error,
```
danawoodman:
```suggestion
        error,
```

backend/src/index.ts
```diff
@@ -26,16 +26,22 @@ Logger.info('Desktop starting up!')
...
 process
   .on('uncaughtException', (error: Error) => {
-    d('Caught exception', error)
-    AirBrake.notify({ message: 'CAUGHT EXCEPTION', error })
-    Logger.warn('CAUGHT EXCEPTION', { error, details: error.toString() })
-    EventBus.emit(CLI.EVENTS.error, error.toString())
+    d('Uncaught exception', error)
+    AirBrake.notify({
+      params: { type: 'UNCAUGHT EXCEPTION' },
+      error: error,
```
danawoodman:
```suggestion
      error,
```

JamieRuderman: fixed

backend/src/Command.ts
JamieRuderman: yes. fixed.

PR #58, opened 2020-02-29
DESK-363 sleep reconnect

JamieRuderman:
Reconnect the UI if disconnected unintentionally.

### DIFF
--- a/frontend/src/pages/Page/Page.tsx
+++ b/frontend/src/pages/Page/Page.tsx
@@ -23,6 +23,7 @@ export function Page({ authenticated = true, children }: Props & React.HTMLProps
   }))
   const css = useStyles()
   const clearCliError = () => backend.set({ key: 'cliError', value: undefined })
+  let message = ''
   let remoteCss = ''
   let pageCss = css.page + ' ' + css.full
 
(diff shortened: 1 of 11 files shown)


frontend/src/pages/Page/Page.tsx
```diff
@@ -32,12 +33,14 @@ export function Page({ authenticated = true, children }: Props & React.HTMLProps
...
 
+  if (authenticated && !connected) message = 'Webserver connection lost. Retrying...'
+
   return (
     <div className={remoteCss}>
       <RemoteHeader os={os} />
       <div className={pageCss}>
         {children}
-        <Snackbar open={authenticated && !connected} message="Webserver connection lost. Retrying..." />
+        <Snackbar open={!!message} message={message} />
```
danawoodman: Minor but I prefer `Boolean(message)` since it is clearer what it means

JamieRuderman: Yea, I still like the !!  - it's just familiarity - like function() {}  vs () => {} or new Array() vs [] or ? : ; vs if then else... 🤷‍♂️

PR #73, opened 2020-03-16
[DESK-387] Auto install cli if run as admin. Unit tests.

JamieRuderman:
### Description

Update to allow cli install without prompting if running as root/admin.
Unit tests.

### Checklist

- [ ] ~~I have prefixed this PR with `[WIP]` if it is a Work In Progress (not ready to merge)~~
- [x] I have added/updated tests for any code changes
- [ ] ~~I have updated documentation (including README, inline comments and docs.remote.it docs)~~
- [ ] ~~Translation strings added/updated for all user-visible text~~

### Test plan

1. Delete cli 
2. Start app as root/admin as described in readme
2. See that cli is now installed without propt

### Ticket

<https://remoteit.atlassian.net/browse/DESK-387>

### DIFF
--- a/backend/src/Installer.ts
+++ b/backend/src/Installer.ts
@@ -1,5 +1,6 @@
 import debug from 'debug'
 import cli from './cliInterface'
+import binaryInstaller from './binaryInstaller'
 import semverCompare from 'semver-compare'
 import environment from './environment'
 import EventBus from './EventBus'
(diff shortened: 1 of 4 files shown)


backend/src/Installer.ts
```diff
@@ -44,13 +45,11 @@ export default class Installer {
...
-  async check() {
+  async check(log?: boolean) {
     d('CHECK INSTALLATION', { name: this.name, version: this.version })
-    const current = await this.isCurrent()
-    current
-      ? EventBus.emit(Installer.EVENTS.installed, this.toJSON())
-      : EventBus.emit(Installer.EVENTS.notInstalled, this.name)
-    return current
+    if (await this.isCurrent(log)) return EventBus.emit(Installer.EVENTS.installed, this.toJSON())
+    if (environment.isElevated) return await binaryInstaller.install()
```
evanrbowers: Seems odd that a check function would do an install.  Recommend breaking into two functions (check and install) or renaming the function.

JamieRuderman: So there is a hearbeat loop that the backend runs and just checks a few things: cli installation, connections and updates. 

Each model has a check function that it runs regularly. If the check needs an action the action is performed.

In terms of the installer check the, actions that it might perform are telling the user that they will be prompted for permission to install the system service, or to just do the installation in the background if they are running as root. 

The connections check will restart a connection that meets the criteria and the update check will notify the user of an available update.

Anyway, since it's a check on the installer I don't feel like it's wrong to have it trigger the installation, it's what it was designed to do. There is another function that is an install function that can be triggered manually. the binaryInstaller.install() method.

Hope that helps!

evanrbowers: That does thanks!

PR #120, opened 2020-04-21
[DESK-489] Implement Segment tracking for analytics

evanrbowers:
### Checklist

- [x] I have updated documentation (including README, inline comments and docs.remote.it docs)

### Ticket

https://remoteit.atlassian.net/projects/DESK/issues/DESK-489


backend/src/constants.ts
```diff
@@ -1,7 +1,10 @@
 import os from 'os'
 import path from 'path'
 
+//General
 export const ENVIRONMENT = process.env.NODE_ENV || 'production'
+export const PRODUCT_NAME = 'Desktop'
+export const MANUFACTURER_NAME = 'Remote.it'
```
danawoodman, line 7: remote.it without capital

backend/src/environment.ts
```diff
@@ -63,13 +67,48 @@ export class Environment {
...
+      try {
+        let versionInfo: any = plist.parse(fs.readFileSync('/System/Library/CoreServices/SystemVersion.plist', 'utf8'))
+        console.log(JSON.stringify(versionInfo.ProductVersion))
+        osVersion = versionInfo.ProductVersion
+      } catch {
+        console.log('No System Version File: /System/Library/CoreServices/SystemVersion.plist')
+      }
+    }
+
+    if (this.manufacturerDetails === undefined) {
```
danawoodman, line 86:
```
if (!this.manufacturerDetails)
```

backend/src/environment.ts
```diff
@@ -63,13 +67,48 @@ export class Environment {
...
+      try {
+        let manufactererFile = JSON.parse(fs.readFileSync(path.join(this.adminPath, 'manufacturer.json'), 'utf8'))
```
danawoodman, line 88: What happens if file doesn't exist? What happens if content isn't in the format we expect?

backend/src/environment.ts
```diff
@@ -63,13 +67,48 @@ export class Environment {
...
+        this.manufacturerDetails = manufactererFile.manufacturer
+      } catch (e) {
+        //catch if file does not exist
+        let version = undefined
+        try {
+          let packageFile = JSON.parse(fs.readFileSync(path.join(process.cwd(), '/package.json'), 'utf8'))
```
danawoodman, line 94: I don't think this is how you want to do it. Can you instead import the package.json file as a normal import?

```
import pkg from '../package.json'
```

common/types.d.ts
```diff
@@ -127,6 +128,12 @@ declare global {
 
   type IConnectionKey = keyof IConnection
 
+  interface IosInfo {
```
danawoodman, line 131: Interfaces should have I prefix:

```
interface OSInfo {
```

common/types.d.ts
```diff
@@ -194,6 +201,12 @@ declare global {
   type ILookup = { [key: string]: any }
 
   type IPreferences = ILookup
+
+  type IManufacturer = {
```
danawoodman, line 205:
```
interface ManufacturerInfo {
```

common/types.d.ts
```diff
@@ -194,6 +201,12 @@ declare global {
...
+    name?: string
+    product: { version?: string; name?: string; code?: number }
```
danawoodman, line 207: Prob should expect product to be optional:

```
  product?: { version?: string; name?: string; code?: number }
```

Same for platform

frontend/src/components/LocalhostScanForm/LocalhostScanForm.tsx
```diff
@@ -26,6 +27,7 @@ export const LocalhostScanForm: React.FC<Props> = ({ setSelected, disabled }) =>
   const css = useStyles()
 
   useEffect(() => {
+    Analytics.Instance.track('ScanLocalNetwork')
```
danawoodman, line 30: Isn't event `scanNetwork`?

frontend/src/helpers/Analytics.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  public setOS(os: any) {
+    this.context.systemOS = os
+  }
+
+  public setOsVersion(version: any) {
+    this.context.systemOSVersion = version
+  }
+
+  public static get Instance() {
```
danawoodman, line 51: This is kinda weird to me. What about just create an instance and export that instead, like:

```js
class Analytics {}

export default new Analytics()
```

Then make all methods not static.

frontend/src/services/Controller.ts
```diff
@@ -129,10 +130,27 @@ function getEventHandlers() {
...
 
+    //Analytics
+    setOSInfo: (osInfo: IosInfo) => {
+      Analytics.Instance.setOS(osInfo.os)
+      Analytics.Instance.setOsVersion(osInfo.version)
+      Analytics.Instance.setArch(osInfo.arch)
+    },
+
     // User
     'signed-out': () => auth.signedOut(),
```
danawoodman, line 155: I would have assumed you'd track things on the backend here. Why was it that you're not?

frontend/src/types.d.ts
```diff
@@ -1,5 +1,7 @@
 import { DeviceState, ServiceState } from 'remote.it'
 
+/// <reference types="@types/segment-analytics" />
```
danawoodman, line 3: Don't use imports like this, instead import the type from the library like a regular import:

```js
import { SomeType } from 'segment'
```

You can see this in latest portal code.

frontend/src/types.d.ts
```diff
@@ -58,6 +60,8 @@ declare global {
     process?: {
       type?: string
     }
+    analytics: SegmentAnalytics.AnalyticsJS
+    ga: S
```
danawoodman, line 64: What is `S`?

backend/src/environment.ts
evanrbowers: Thats how I did it originally.  It did not work.  works in other files I have tried but not this one.  If you find a way to make it work let know.

frontend/src/buttons/ConnectButton/ConnectButton.tsx
```diff
@@ -21,7 +22,17 @@ export const ConnectButton: React.FC<ConnectButtonProps> = ({
 }) => {
   const hidden = (connection && connection.active) || !service || service.state !== 'active'
   const connecting = !!(connection && connection.pid && !connection.active)
-  const connect = () => emit('service/connect', connection || newConnection(service))
+  const connect = () => {
+    let theConnection = connection || newConnection(service)
+    let context = {
+      connectionType: 'peer',
+      serviceId: service?.deviceID,
```
JamieRuderman, line 29: There is a service id as well as a device id... the serviceId is not the DeviceId.

backend/src/constants.ts
evanrbowers: updated

backend/src/environment.ts
evanrbowers: updated

frontend/src/buttons/DisconnectButton/DisconnectButton.tsx
```diff
@@ -27,7 +31,16 @@ export const DisconnectButton: React.FC<Props> = ({
           color={color}
           disabled={disabled}
           size={size}
-          onClick={() => emit('service/disconnect', connection)}
+          onClick={() => {
+            let context = {
+              connectionType: 'peer',
+              serviceId: service?.deviceID,
```
JamieRuderman, line 37: Same here `serviceId !== deviceId`

frontend/src/helpers/Analytics.ts
JamieRuderman: Yes please.

backend/src/environment.ts
evanrbowers: then the catch occurs and we load defaults

common/types.d.ts
evanrbowers: I'm not sure what you are saying.  it does have an I prefix.

frontend/src/pages/DebugPage/DebugPage.tsx
```diff
@@ -1,9 +1,14 @@
-import React from 'react'
+import React, { useEffect } from 'react'
 import { DebugLogController } from '../../controllers/DebugLogController'
 import { Typography } from '@material-ui/core'
 import { Page } from '../Page'
+import Analytics from '../../helpers/Analytics'
 
 export function DebugPage() {
+  useEffect(() => {
```
JamieRuderman, line 8: I think maybe a more elegant solution to page tracking would be to provide a custom useHistory hook... maybe `useAndTrackHistory` where we just track the react router history changes. Then we don't have to have this useEffect on every page load.

common/types.d.ts
evanrbowers: Do you want this to be IManufacturerInfo?

frontend/src/components/LocalhostScanForm/LocalhostScanForm.tsx
evanrbowers: yeah, should be networkScan. missed the second one. updated

frontend/src/helpers/Analytics.ts
evanrbowers: I could be wrong but to me that seems like that would create a new Analytics object every time you imported it.

frontend/src/services/Controller.ts
evanrbowers: some things had to be done in the frontend.  It was easier to implement in one place then doing it in both.  This particular code would be needed for all the frontend calls regardless.

frontend/src/types.d.ts
evanrbowers: left over from some initial trials.  Removed.

common/types.d.ts
evanrbowers: updated

backend/src/environment.ts
```diff
@@ -63,13 +67,48 @@ export class Environment {
...
 
   get frontend() {
+    let osVersion = os.release()
+    if (this.isMac) {
+      try {
+        let versionInfo: any = plist.parse(fs.readFileSync('/System/Library/CoreServices/SystemVersion.plist', 'utf8'))
+        console.log(JSON.stringify(versionInfo.ProductVersion))
+        osVersion = versionInfo.ProductVersion
+      } catch {
+        console.log('No System Version File: /System/Library/CoreServices/SystemVersion.plist')
```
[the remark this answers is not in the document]
evanrbowers: fixed thanks!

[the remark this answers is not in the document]
evanrbowers: good catch, did this one earlier before I realized how it was all hung together =).  Moved.

common/types.d.ts
JamieRuderman: He is saying that new thinking is to not Prepend the `I` but everything is already with the `I` so maybe at some point we can convert everything. I'm not sure it's super high priority.

frontend/src/helpers/Analytics.ts
JamieRuderman: No, this is the standard way to create a Singleton instance of a class. It will only be executed once.

frontend/src/services/Controller.ts
JamieRuderman: I think this is fine. It is very light weight, and already being passed to the frontend.

frontend/public/index.html
```diff
@@ -8,6 +8,58 @@
     <meta name="viewport" content="minimum-scale=1, initial-scale=1, width=device-width, shrink-to-fit=no" />
     <link rel="manifest" href="%PUBLIC_URL%/manifest.json" />
     <title>remote.it</title>
+    <script>
+      !(function () {
```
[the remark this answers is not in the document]
evanrbowers: We implemented your suggestion =)

frontend/src/buttons/ConnectButton/ConnectButton.tsx
evanrbowers: change to .id

frontend/src/buttons/DisconnectButton/DisconnectButton.tsx
evanrbowers: change to .id

frontend/src/pages/DebugPage/DebugPage.tsx
evanrbowers: I like this idea if we can get the page name.  lets talk

backend/src/constants.ts
danawoodman: I don't see any changes, sure you pushed?

backend/src/environment.ts
danawoodman: @JamieRuderman He will probably have to do this at the top of the file right?

common/types.d.ts
danawoodman: Should not have an "I" prefix sorry, was a typo

backend/src/constants.ts
evanrbowers: pushed

backend/src/environment.ts
evanrbowers: Import did not work at the top of the file.

frontend/src/helpers/Analytics.ts
evanrbowers: updated with this method but it did cause some issues with initialization.  now must call window.analytics directly from inside Analytics class rather then using an object variable.  Jamie saw this issue with me.

frontend/src/pages/DebugPage/DebugPage.tsx
evanrbowers: we decided to do this later as it would be a large change and works as is for now.

evanrbowers:
> Some minor suggests.
> 
> Did you version control lock files?

lock files checked in

evanrbowers:
> Update backend/package.json & frontend/package.json & package.json with version 2.5.24-alpha.9
> some file conflicts with the last commit from Jamie

version updated and merge conflicts resolved

backend/src/environment.ts
JamieRuderman: Yea, maybe it's just needs a typescript restart... not sure. Should work.

PR #124, opened 2020-04-23
[DESK-522] protect against segment errors

evanrbowers:
### Description

We are getting some segment errors in the field.  We need to wrap the segment calls in a try catch to keep them from interrupting the user.

### Checklist

### Test plan

1. Sign in to segments.io
2. Launch Desktop
3. make a connection
4. Verify page view and connection attempt are logged in segment and no errors occures

### Ticket

https://remoteit.atlassian.net/browse/DESK-522

### DIFF
--- a/backend/package.json
+++ b/backend/package.json
@@ -1,6 +1,6 @@
 {
   "name": "remoteit-headless",
-  "version": "2.5.24-alpha.10",
+  "version": "2.5.24-alpha.11",
   "private": true,
   "main": "build/index.js",
   "scripts": {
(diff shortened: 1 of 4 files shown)


backend/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "remoteit-headless",
-  "version": "2.5.24-alpha.10",
+  "version": "2.5.24-alpha.11",
```
JamieRuderman, line 3: You don't need to increment every commit. I'll increment and tag right before I do a build.

evanrbowers:
I agree with Jamie. This issue is actually coming from CLI and this code is not needed.

PR #141, opened 2020-05-07
[DESK-542] Application level state

JamieRuderman:
### Description

Moved page state in scanning and device setup to the application level, so that it's retained across page changes and should be protected from users accidentally or mistakenly seeing wrong state.

### Checklist

- [ ] ~~I have prefixed this PR with `[WIP]` if it is a Work In Progress (not ready to merge)~~
- [ ] ~~I have added/updated tests for any code changes~~
- [ ] ~~I have updated documentation (including README, inline comments and docs.remote.it docs)~~
- [ ] ~~Translation strings added/updated for all user-visible text~~

### Test plan

1. Sign in
2. Register a device, or use an already registered device
3. Add and remove services manually and with the network scan
4. Change pages while in progress
5. The state should be retained even when going to a new page and returning

### Ticket

<https://remoteit.atlassian.net/browse/DESK-542>

### DIFF
--- a/frontend/src/models/ui.ts
+++ b/frontend/src/models/ui.ts
@@ -9,6 +9,9 @@ type UIState = UIParams & {
   scanLoading: { [interfaceName: string]: boolean }
   scanTimestamp: { [interfaceName: string]: number }
   scanInterface: string
+  setupDeletingService: boolean
+  setupServicesCount: number
+  setupServicesNew: boolean
 }
 
 const state: UIState = {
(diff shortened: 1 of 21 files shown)


frontend/src/models/ui.ts
```diff
@@ -1,29 +1,57 @@
...
+  scanLoading: {},
+  scanTimestamp: {},
+  scanInterface: DEFAULT_INTERFACE,
+  setupBusy: false,
+  setupAdded: undefined,
+  setupDeletingDevice: false,
+  setupDeletingService: undefined,
+  setupAddingService: false,
+  setupServicesCount: 0,
+  setupServicesNew: true,
```
danawoodman, line 33: Why not a nested object for this instead? Feels a bit clunky

JamieRuderman: TBH I wasn't sure that redux would trigger a change event on just the one item that was changed and not unnecessarily trigger a change event on the other items if they were nested.

PR #217, opened 2020-08-24
DESK-412 open windows terminal

mauricio0000:
### Checklist

- [ ] Pull Request title is in the format `[PROJ-12345] Some useful description here...`
- [ ] I have prefixed this PR with `[WIP]` if it is a Work In Progress (not ready to merge)
- [ ] The git branch is in the format `PROJ-12345-some-useful-description-here`
- [ ] I have added one or more reviewers
- [ ] I have added/updated tests for any code changes
- [ ] I have updated documentation (including README, inline comments and docs.remote.it docs)
- [ ] This PR only contains one isolated change (bug fix, feature, etc)
- [ ] Translation strings added/updated for all user-visible text
- [ ] I have manually reviewed this PR to make sure only the files I intended to change were changed and nothing else
- [ ] I have notified reviewers on Slack by @ mentioning them in the `#desktop`channel
- [ ] I have moved the Jira ticket into `Resolved` state and made a note in the ticket with a link to this PR

### Test plan

1. Sign in
2. Go to...
3. Click...
4. You should see X, Y, Z...

### Ticket

<https://remoteit.atlassian.net/browse/TICKET_NUMBER_HERE>


backend/src/ConnectionPool.ts
```diff
@@ -190,4 +191,9 @@ export default class ConnectionPool {
       return result
     }, [])
   }
+
+  openCMDforWindows(launchApp: string) {
+    Logger.error(launchApp)
+    child_process.exec(`start cmd.exe /K ${launchApp}`)
+  }
```
JamieRuderman, line 198: Let's make a new file for this feature. Maybe call it `launch.ts`. I don't think it belongs in the connection pool class.

common/types.d.ts
```diff
@@ -27,6 +27,7 @@ declare global {
     | 'service/forget'
     | 'service/restart'
     | 'service/clear-recent'
+    | 'service/cmd/open'
```
JamieRuderman, line 30: How about we just call this `service/launch`

frontend/src/buttons/LaunchButton/LaunchButton.tsx
```diff
@@ -42,7 +44,10 @@ export const LaunchButton: React.FC<Props> = ({ connection, service }) => {
         ...connection,
         username: username.toString(),
       })
-    window.open(app.launch({ ...connection, username }))
+    const launchApp = app.launch({ ...connection, username })
+    console.log({app, launchApp})
+    isWindows() && app.title === 'SSH' ?
```
JamieRuderman, line 49: Can we move this logic into the `/shared/applications.ts` file?

Please add a new method on the Application class that returns a boolean if it should launch to the browser.

So on the base class the method could return by default `true`.

Then the instance of 'SSH' should have an override where it checks for isWindows().

Let me know if that all makes sense... I'm happy to jump on a call!  :)

backend/src/sharedCopy/applications.ts
```diff
@@ -39,6 +39,11 @@ class Application {
     const matches: string[] = result.match(/\[[^\]]*\]/g) || []
     return matches.map(m => m.slice(1, -1))
   }
+
+  launchBrowser(type: string) {
+    const platform = navigator ? navigator.platform.toLowerCase() : ''
+    return !(type === 'SSH' && platform.includes('win'))
```
JamieRuderman, line 45: There is a `Browser.ts` service file that has an isWindows() method already you can use here.

```suggestion
    return !(type === 'SSH' && isWindows())
```

frontend/src/buttons/LaunchButton/LaunchButton.tsx
```diff
@@ -42,16 +51,29 @@ export const LaunchButton: React.FC<Props> = ({ connection, service }) => {
         ...connection,
         username: username.toString(),
       })
-    window.open(app.launch({ ...connection, username }))
+    const launchApp = app.launch({ ...connection, username })
+
+    app.launchBrowser(app.title) && false ? window.open(launchApp) : emit('service/launch', launchApp)
```
JamieRuderman, line 56:
```suggestion
    app.launchBrowser(app.title) ? window.open(launchApp) : emit('service/launch', launchApp)
```
Looks like debug code here

backend/src/sharedCopy/applications.ts
mauricio0000: `isWindows()` is going to broke the copyShare in backend

PR #241, opened 2020-09-09
[DESK-742] device list no name

JamieRuderman:
### Description

Remove the name field from graphQL device list query if empty

### Ticket

<https://remoteit.atlassian.net/browse/[DESK-742]>

[DESK-742]: https://remoteit.atlassian.net/browse/DESK-742


penngrove (changes requested):
Do not change the query, but do not pass the parameter name

penngrove (approved):

PR #262, opened 2020-10-01
[DESK-727] ssh copy username

mauricio0000:
### Test plan

1. Sign in
2. Go to one device
3. Connect to a service SSH
4. Set up the username
5. Copy the command to connect
6. You should see the username on the clipboard

### Ticket

https://remoteit.atlassian.net/browse/DESK-727


JamieRuderman (commented):
What would you think about breaking out the dialog into it's own component and share it with the ssh launch component?

frontend/src/buttons/CopyButton/CopyButton.tsx
```diff
@@ -16,8 +18,29 @@ export const CopyButton: React.FC<CopyButtonProps> = ({ connection, service, men
   const clipboard = useClipboard({ copiedTimeout: 1000 })
   const app = useApplication(service && service.typeID)
 
+  const [username, setUsername] = useState<string>((connection && connection.username) || '')
```
JamieRuderman, line 21: Let's move username into the modal. 
```suggestion
```

frontend/src/components/ModalSetUsername/ModalSetUsername.tsx
```diff
@@ -0,0 +1,50 @@
...
+import useApplication from '../../shared/applications'
+
+export const ModalSetUsername: React.FC<{
+  connection?: IConnection
+  openUsername: boolean
+  handleSubmit: () => void
+  username: string
+  setUsername: React.Dispatch<React.SetStateAction<string>>
+  service?: IService
+  close: () => void
```
JamieRuderman, line 12:
```suggestion
  connection?: IConnection
  open: boolean
  service?: IService
  onSubmit: (username: string) => void
  onClose: () => void
```
Let's have the modal keep the username state and return the username when it's submitted.
I don't think we need to say username again on the open because we know it's the username modal.
And I like to follow the material UI convention of prefixing any callbacks with "on" so it becomes onClose and onSubmit.

frontend/src/components/ModalSetUsername/ModalSetUsername.tsx
```diff
@@ -0,0 +1,50 @@
+import { Button, Dialog, DialogActions, DialogContent, TextField, Typography } from '@material-ui/core'
+import React from 'react'
+import useApplication from '../../shared/applications'
+
+export const ModalSetUsername: React.FC<{
```
JamieRuderman, line 5: Could we call this component UsernameModal? It follows the pattern of <descriptorType><componentType> like LaunchButton or CopyButton
```suggestion
export const UsernameModal: React.FC<{
```

JamieRuderman (changes requested):
Looking good just some suggestions!

JamieRuderman (approved):
Looks great thank you! 
Please wait for me to merge future to master to merge.

PR #266, opened 2020-10-05
[DESK-771][DESK-777] account linking blank screen

JamieRuderman:
### Description

device list sharing
registration workflow
refreshing single devices
viewing shared device list connections
Fix for graphQL error when device isn't registered

### Test plan

1. Sign in
2. Go to...
3. Click...
4. You should see X, Y, Z...

### Ticket

<https://remoteit.atlassian.net/browse/DESK-771>
<https://remoteit.atlassian.net/browse/DESK-777>


frontend/src/helpers/connectionHelper.ts
```diff
@@ -49,6 +51,13 @@ export function clearConnectionError(connection: IConnection) {
   emit('connection', connection)
 }
 
+export function getConnectionIds(state: ApplicationState) {
+  const { device, connections } = state.backend
+  let ids = connections.map(c => c.id)
```
mauricio0000, line 56: why do you need a `let`?

frontend/src/helpers/connectionHelper.ts
```diff
@@ -49,6 +51,13 @@ export function clearConnectionError(connection: IConnection) {
...
+  if (device.uid) ids.push(device.uid)
```
mauricio0000, line 57: could you add validation to check if the id is not duplicate? I am not sure

backend/src/Controller.ts
```diff
@@ -161,8 +160,7 @@ class Controller {
 
   clearAll = async () => {
     Logger.info('CLEAR CREDENTIALS')
-    await user.clearAll() // moved to before clearing connections because having many connection would cause sign out to fail.
-    await this.pool.clearAll()
+    await user.clearAll()
```
mauricio0000, line 163: why do you need `await`?

frontend/src/models/devices.ts
```diff
@@ -97,38 +92,37 @@ export default createModel({
...
       try {
-        const gqlResponse = await graphQLFetchDevice(deviceId, accountId)
+        const gqlResponse = await graphQLFetchDevice(deviceId)
         graphQLGetErrors(gqlResponse)
-        const { device } = gqlResponse?.data?.data?.login?.account || {}
-        result = graphQLAdaptor([device], accountId)[0]
+        const { device } = gqlResponse?.data?.data?.login || {}
+        const loginId = gqlResponse?.data?.data?.login?.id
+        result = device ? graphQLAdaptor(device, loginId, hidden)[0] : []
+        debugger
```
mauricio0000, line 109: could you remove the `debugger`?

mauricio0000 (changes requested):

frontend/src/helpers/connectionHelper.ts
JamieRuderman: well I'm pushing to the array in the next line, so I assume it's gotta be let or var... not const if it's changing.  🤷‍♂️

frontend/src/helpers/connectionHelper.ts
JamieRuderman: Sure why not!

backend/src/Controller.ts
JamieRuderman: No, don't need await... and this is actually changing again... I'll update...

frontend/src/models/devices.ts
JamieRuderman: Yes. Definitely.

mauricio0000 (approved):

PR #273, opened 2020-10-07
[DESK-0000]fix fetch device-after delete user

mauricio0000:
### Description

[DESK-0000]fix fetch device-after delete user


JamieRuderman (approved):
Thank you !

PR #281, opened 2020-10-12
[DESK-457] Bundling binaries

mauricio0000:
### Description

[DESK-457] build binaries

### Test plan

1. Build a new version for windows
2. Install
3. You should see CLI installed using the resource

### Ticket
https://remoteit.atlassian.net/browse/DESK-457

[DESK-457]: https://remoteit.atlassian.net/browse/DESK-457


JamieRuderman (commented):
Looking great, but what about updating the installation through binaryInstaller.ts? Let's discuss when we should merge.

backend/src/Installer.ts
```diff
@@ -126,14 +151,23 @@ export default class Installer {
    * Download the binary, move it to the PATH on the user's
    * system and then make it writable.
    */
-  install(tempDir: string, cb?: ProgressCallback) {
+  install(cb?: ProgressCallback) {
```
JamieRuderman, line 154: This method used to download and move the files... I would think since the files are already local that we would just be selecting and moving the files and not need a tmp dir? Probably will have to refactor more and update the installBinaries.ts file as well.

backend/src/Installer.test.ts
```diff
@@ -9,6 +9,12 @@ import fs from 'fs'
 describe('backend/Installer', () => {
   describe('check', () => {
     const version = '0.37.6'
+    const versionMuxer = '0.37.6'
+    const versionDemuxer = '0.37.6'
+    const versionConnectd = '0.37.6'
+    const baseUrl = 'https://downloads.remote.it/cli/v'
+    const cliUrl = 'https://downloads.remote.it/cli/v'
+    const connectdUrl= 'https://downloads.remote.it/cli/v'
```
JamieRuderman, line 17: Instead of handling each of the four assets differently lets just put them in a `resources` or `files` array and then loop over them. Since there is more than a version number and name (and might have more info) it could look something like this?:
```
const resources = [{
  name: 'remoteit',
  version: '6.1.14'
}, {
  name: 'connectd',
  version: '4.9'
} , {
  ...
}]
```
Actually the `installer.ts` file was designed to handle a single file install, and the `binaryInstaller.ts` was designed to iterate over instances of an installer to install each file. We could still do something like that.

backend/src/Installer.ts
```diff
@@ -152,71 +186,129 @@ export default class Installer {
...
+
+  get downloadFileNameConnectd() {
+    let platform = ''
+    if (environment.isWindows32) platform = 'connectd.x86-win.exe'
+    else if (environment.isWindows) platform = 'connectd.x86_64-64.exe'
+    else if (environment.isMac) platform = 'connectd.x86_64-osx'
+    return platform
+  }
+
+  binaryPathCLI(admin?: boolean) {
```
JamieRuderman, line 213: Yea we can do this without repeating ourselves. Wanna keep the code ad DRY as possible.  🙂

backend/src/binaryInstaller.test.ts
```diff
@@ -23,7 +23,7 @@ describe('backend/binaryInstaller', () => {
       commandSpy = jest.spyOn(Command.prototype, 'push').mockImplementation()
       downloadSpy = jest
         .spyOn(binaryInstaller, 'download')
-        .mockImplementation((i: Installer, t: tmp.DirResult) => Promise.resolve())
+        // .mockImplementation((i: Installer, t: tmp.DirResult) => Promise.resolve())
```
JamieRuderman, line 26: Yea once downloading is gone this should be ripped out and cleaned up

backend/src/binaryInstaller.ts
```diff
@@ -40,33 +40,36 @@ class BinaryInstaller {
...
       const commands = new Command({ onError: reject, admin: true })
 
       if (environment.isWindows) {
         if (!existsSync(environment.binPath)) commands.push(`md "${environment.binPath}"`)
-        commands.push(`move /y "${installer.tempFile}" "${installer.binaryPath()}"`)
-        commands.push(`icacls "${installer.binaryPath()}" /T /C /Q /grant "*S-1-5-32-545:RX"`) // Grant all group "Users" read and execute permissions
+        commands.push(`move /y "${installer.tempFile}" "${installer.binaryPathCLI()}"`)
+        commands.push(`move /y "${installer.tempFileMuxer}" "${installer.binaryPathMuxer()}"`)
+        commands.push(`move /y "${installer.tempFileDemuxer}" "${installer.binaryPathDemuxer()}"`)
+        commands.push(`move /y "${installer.tempFileConnectd}" "${installer.binaryPathConnectd()}"`)
```
JamieRuderman, line 56: So instead of moving these, let's leave them in their source directory and just `ln -l` them into the installation location.

backend/src/constants.ts
```diff
@@ -61,13 +68,13 @@ export const PATHS = {
 
   MAC_USER_BINARIES: path.join(os.homedir(), '.remoteit/bin'),
   MAC_USER_SETTINGS: path.join(os.homedir(), '.remoteit'),
-  MAC_ADMIN_BINARIES: '/usr/local/bin',
+  MAC_ADMIN_BINARIES: './resources/mac',
```
JamieRuderman, line 71: Could this path be the same on both mac and win?
```suggestion
  MAC_ADMIN_BINARIES: './resources',
```

backend/src/remoteitInstaller.ts
```diff
@@ -1,9 +1,23 @@
...
-import { CLI_VERSION } from './constants'
+import {
+  BASE_URL,
+  CLI_URL,
+  CLI_VERSION,
+  CONNECTD_URL,
+  CONNECTD_VERSION,
+  DEMUXER_VERSION,
+  MUXER_VERSION,
+} from './constants'
```
JamieRuderman, line 10: This is the right file to make the array of installers - where each one represents a binary to install... [remoteit, connectd, muxer, demuxer]...

installer.nsh
```diff
@@ -104,6 +62,11 @@
         ${EndIf}
         FileWrite $installLog "- $REMOTE_CLI_EXE -j service install     [$0]  $1$\r$\n"
 
+        CopyFiles $REMOTE_CLI_EXE $PATH_REMOTE_DIR
+        CopyFiles $MUXER $PATH_REMOTE_DIR
+        CopyFiles $DEMUXER $PATH_REMOTE_DIR
+        CopyFiles $CONNECTD $PATH_REMOTE_DIR
+
```
JamieRuderman, line 69: Want to set path var instead of copying.

https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_environment_variables?view=powershell-7

JamieRuderman (commented):
Looking great. Lots of progress. Please see my notes. Happy to work with you tomorrow if you'd like. Thanks!

backend/src/Installer.test.ts
```diff
@@ -21,13 +22,17 @@ describe('backend/Installer', () => {
...
     beforeAll(() => {
       installer = new Installer({
-        name,
-        version,
+        resources: [
+          { name: 'remoteit', version: version.cli, url: version.url_cli },
+          { name: 'muxer', version: version.muxer, url: version.url_base },
+          { name: 'demuxer', version: version.demuxer, url: version.url_base },
+          { name: 'connectd', version: version.connectd, url: version.url_connectd },
+        ],
```
JamieRuderman, line 30: Not sure it matters, but why duplicate the resources that are in the constants here?

backend/src/Installer.ts
```diff
@@ -64,25 +63,26 @@ export default class Installer {
...
-    const check = this.dependencyNames.concat(this.binaryName)
-    const missing = check.find(fileName => !this.fileExists(fileName))
-    d('IS INSTALLED?', { installed: !missing })
-    return !missing
+    return true
+    // if (binaryInstaller.inProgress) return false
+    // const check = this.dependencyNames.concat(this.binaryName)
+    // const missing = check.find(fileName => !this.fileExists(fileName))
+    // d('IS INSTALLED?', { installed: !missing })
+    // return !missing
```
JamieRuderman, line 71: We should clean up this commented out stuff - just remove the stuff that isn't used any more.

backend/src/Installer.ts
```diff
@@ -152,71 +160,123 @@ export default class Installer {
...
-            stream(res1)
-          } else {
-            Logger.warn('No download header in download URL', { headers: res1.headers })
-            return reject(new Error('No download header in download URL!'))
+  //     this.getFileWeb(url, resolve, reject, progress, this.binaryPathCLI())
+  //     this.getFileWeb(muxer, resolve, reject, progress, this.binaryPathMuxer())
+  //     this.getFileWeb(demuxer, resolve, reject, progress, this.binaryPathDemuxer())
+  //     this.getFileWeb(connectd, resolve, reject, progress, this.binaryPathConnectd())
+  //   })
+  // }
```
JamieRuderman, line 235: Just remove this stuff

backend/src/Installer.ts
```diff
@@ -152,71 +160,123 @@ export default class Installer {
...
+
+  get downloadFileNameDemuxer() {
+    return `demuxer${this.extensionName}`
+  }
+
+  get downloadFileNameConnectd() {
+    return `connectd${this.extensionName}`
+  }
+
+  binaryPathCLI(admin?: boolean) {
```
JamieRuderman, line 183: So the Installer.ts is meant to install a single dependancy. So one installer for each resource. We shouldn't add multi resources to the installer, instead create a new instance of the installer for each resource.

backend/src/Installer.ts
```diff
@@ -11,25 +10,26 @@ import https from 'https'
 import fs from 'fs'
 import path from 'path'
 import { existsSync } from 'fs'
+import tmp from 'tmp'
+
+tmp.setGracefulCleanup()
```
JamieRuderman, line 15: I don't see this being used... maybe I missed it.

backend/src/Installer.ts
```diff
@@ -100,16 +100,26 @@ export default class Installer {
...
+          cliVersion,
+          desiredVersion: this.resources[0].version,
+        })
     } else {
-      Logger.info('CLI NOT CURRENT', { name: this.name, cliVersion, desiredVersion: this.version })
+      Logger.info('CLI NOT CURRENT', {
+        name: this.resources[0].name,
+        cliVersion,
+        desiredVersion: this.resources[0].version,
+      })
```
JamieRuderman, line 122: Wont need these changes if you make a single resource per instance.

backend/src/Installer.ts
```diff
@@ -152,71 +160,123 @@ export default class Installer {
...
   }
 
-  binaryPath(admin?: boolean) {
+  get extensionName() {
+    let platform = ''
+    if (environment.isWindows32) platform = '.x86-win.exe'
+    else if (environment.isWindows) platform = '.x86_64-64.exe'
+    else if (environment.isMac) platform = '.x86_64-osx'
+    return platform
+  }
```
JamieRuderman, line 169: We shouldn't need any of this stuff related to downloading, as we won't be downloading any more.

backend/src/binaryInstaller.ts
```diff
@@ -40,33 +41,28 @@ class BinaryInstaller {
...
   async installBinary(installer: Installer) {
     return new Promise(async (resolve, reject) => {
-      var tmpDir = tmp.dirSync({ unsafeCleanup: true, keep: true })
-
       // Stop and remove old binaries
-      await this.migrateBinaries(installer.binaryPath())
-
-      // Download and install binaries
-      await this.download(installer, tmpDir)
+      // await this.migrateBinaries(installer.binaryPathCLI())
```
JamieRuderman, line 45: We will want to put in some cleanup commands here to remove old binaries that are located in /Program Files/remoteit-bin/ or in /usr/local/bin or /usr/bin etc... that's what this line was doing, so we'll want to bring it back and update it.

backend/src/binaryInstaller.ts
```diff
@@ -40,33 +41,28 @@ class BinaryInstaller {
...
 
       const commands = new Command({ onError: reject, admin: true })
 
       if (environment.isWindows) {
         if (!existsSync(environment.binPath)) commands.push(`md "${environment.binPath}"`)
-        commands.push(`move /y "${installer.tempFile}" "${installer.binaryPath()}"`)
-        commands.push(`icacls "${installer.binaryPath()}" /T /C /Q /grant "*S-1-5-32-545:RX"`) // Grant all group "Users" read and execute permissions
+        // commands.push(`icacls "${installer.binaryPathCLI()}" /T /C /Q /grant "*S-1-5-32-545:RX"`) // Grant all group "Users" read and execute permissions
```
JamieRuderman, line 51: We need to add 'remoteit' to the PATH in windows here.

backend/src/binaryInstaller.ts
```diff
@@ -106,10 +102,10 @@ class BinaryInstaller {
...
   }
 
-  async download(installer: Installer, tmpDir: tmp.DirResult) {
-    return installer
-      .install(tmpDir.name, (progress: number) => EventBus.emit(Installer.EVENTS.progress, { progress, installer }))
-      .catch(error => EventBus.emit(Installer.EVENTS.error, error.message))
+  async download(installer: Installer) {
+    // return installer
+    //   .install((progress: number) => EventBus.emit(Installer.EVENTS.progress, { progress, installer }))
+    //   .catch(error => EventBus.emit(Installer.EVENTS.error, error.message))
```
JamieRuderman, line 108: Remove and cleanup.

backend/src/binaryInstaller.ts
```diff
@@ -124,7 +120,7 @@ class BinaryInstaller {
       const options = { disableGlob: true }
 
       try {
-        rimraf.sync(installer.binaryPath(), options)
+        rimraf.sync(installer.binaryPathCLI(), options)
```
JamieRuderman, line 123: Uninstall should also loop over the installer resource instances and remove the symlinks for each.

backend/src/binaryInstaller.ts
```diff
@@ -40,33 +41,28 @@ class BinaryInstaller {
...
       if (environment.isWindows) {
         if (!existsSync(environment.binPath)) commands.push(`md "${environment.binPath}"`)
-        commands.push(`move /y "${installer.tempFile}" "${installer.binaryPath()}"`)
-        commands.push(`icacls "${installer.binaryPath()}" /T /C /Q /grant "*S-1-5-32-545:RX"`) // Grant all group "Users" read and execute permissions
+        // commands.push(`icacls "${installer.binaryPathCLI()}" /T /C /Q /grant "*S-1-5-32-545:RX"`) // Grant all group "Users" read and execute permissions
       } else {
-        if (!existsSync(environment.binPath)) commands.push(`mkdir -p ${environment.binPath}`)
-        commands.push(`mv ${installer.tempFile} ${installer.binaryPath()}`)
-        commands.push(`chmod 755 ${installer.binaryPath()}`) // @TODO if this is going in the user folder must have user permissions
+        commands.push(`ln -sf ${installer.binaryPathCLI()} /usr/local/bin/`)
```
JamieRuderman, line 53: This is where we loop over the resource instances.

backend/src/remoteitInstaller.ts
```diff
@@ -1,9 +1,8 @@
 import Installer from './Installer'
-import { CLI_VERSION } from './constants'
+import { RESOURCES } from './constants'
 
 export default new Installer({
```
JamieRuderman, line 4: @mauricio0000 This is the place where you should create a `new Installer()` for each resource.
Loop over the `RESOURCES` array and create an installer for each. You might just move the whole RESOURCES constant to this file. 

Then rename this file `installers.ts` and export the array of installer instances for the resources.

Please let me know if you have any questions I'm happy to pair with you on this. Thanks!

installer.nsh
```diff
@@ -20,69 +24,23 @@
...
-    FileClose $4
 
-    Var /GLOBAL PLATFORM
     ${If} ${RunningX64}
-        StrCpy $PLATFORM "remoteit_windows_x86_64.exe"
         FileWrite $installLog "- Platform X64$\r$\n"
+        StrCpy $REMOTE_CLI_EXE "$BINARIES\x64\remoteit.exe"
+        StrCpy $MUXER "$BINARIES\x64\muxer.exe"
+        StrCpy $DEMUXER "$BINARIES\x64\demuxer.exe"
+        StrCpy $CONNECTD "$BINARIES\x64\connectd.exe"
```
JamieRuderman, line 33: We don't want to copy anything here, just add to the windows PATH so that the `remoteit` command will work without the full path.

installer.nsh
```diff
@@ -20,69 +24,23 @@
...
     ${Else}
-        StrCpy $PLATFORM "remoteit_windows_x86.exe"
         FileWrite $installLog "- Platform X86$\r$\n"
+        StrCpy $REMOTE_CLI_EXE "$BINARIES\x86\remoteit.exe"
+        StrCpy $MUXER "$BINARIES\x86\muxer.exe"
+        StrCpy $DEMUXER "$BINARIES\x86\demuxer.exe"
+        StrCpy $CONNECTD "$BINARIES\x86\connectd.exe"
```
JamieRuderman, line 39: Same as above

installer.nsh
JamieRuderman: See above link. No copying.

installer.nsh
```diff
@@ -140,6 +103,8 @@
             ; MessageBox MB_OK "This is a UPDATE!" 
             FileWrite $uninstallLog "$\nUpdate (${__DATE__} ${__TIME__}): $\r$\n"
             FileWrite $uninstallLog "-----------------------------$\r$\n"
+            RMDir /r "$PATH_REMOTE_DIR_U"
+            FileWrite $uninstallLog "- RMDir $PATH_REMOTE_DIR_U$\r$\n"
```
JamieRuderman, line 107: Wont need these. Instead it will be removing from PATH

resources/.gitignore
```diff
@@ -0,0 +1,4 @@
+connectd
+demuxer
+muxer
+remoteit
```
JamieRuderman, line 4: I'd rather keep all the .gitignore stuff in the single file if possible please  :)

JamieRuderman (changes requested):
@mauricio0000 So glad it's working, now just for the cleanup and adding to the PATH on windows.
Also I think I messed up the PR by merging in that branch, so you might want to re-merge in future or something because other commits are appearing. Thanks!

installer.nsh
```diff
@@ -224,20 +170,18 @@
...
+            FileWrite $uninstallLog "- RMDir $INSTDIR$\r$\n"
 
-    RMDir /r "$PROFILE\AppData\Local\remoteit"
-    FileWrite $uninstallLog "- RMDir $PROFILE\AppData\Local\remoteit$\r$\n"
+            FileWrite $uninstallLog "$\n***** End Uninstall ******$\r$\n"
+            FileClose $uninstallLog 
 
-    RMDir /r "$INSTDIR"
-    FileWrite $uninstallLog "- RMDir $INSTDIR$\r$\n"
+        ${endif}
```
JamieRuderman, line 185: Still need to add the the PATH here for windows.

backend/src/binaryInstaller.ts
```diff
@@ -40,33 +40,30 @@ class BinaryInstaller {
...
-      await this.download(installer, tmpDir)
+      await this.migrateBinaries(installer.binaryPathCLI())
 
       const commands = new Command({ onError: reject, admin: true })
 
       if (environment.isWindows) {
         if (!existsSync(environment.binPath)) commands.push(`md "${environment.binPath}"`)
-        commands.push(`move /y "${installer.tempFile}" "${installer.binaryPath()}"`)
-        commands.push(`icacls "${installer.binaryPath()}" /T /C /Q /grant "*S-1-5-32-545:RX"`) // Grant all group "Users" read and execute permissions
+        commands.push(`icacls "${installer.binaryPathCLI()}" /T /C /Q /grant "*S-1-5-32-545:RX"`)
```
JamieRuderman, line 50: Need to add the the system PATH here.

JamieRuderman (changes requested):
Looking better, still some things to do... Also looks like the unit tests aren't passing in circle ci.

backend/src/CLI.ts
```diff
@@ -247,8 +247,12 @@ export default class CLI {
 
     let result
     let commands = new Command({ admin, quiet })
+    if (environment.isWindows) {
+      cmds.forEach(cmd => commands.push(`"${remoteitInstaller.binaryPathCLI()}" ${cmd}`))
```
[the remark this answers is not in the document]
JamieRuderman: Or maybe it has to say remoteit.exe?

backend/src/Installer.ts
```diff
@@ -1,35 +1,30 @@
...
 import { existsSync } from 'fs'
+import binaryInstaller from './binaryInstaller'
+import EventBus from './EventBus'
 
 const d = debug('installer')
 
 export type ProgressCallback = (percent: number) => void
 
 interface InstallerArgs {
-  name: string
```
JamieRuderman, line 20: Please continue to use this.name and this.version instead of using the resourced definitions. I think we might be able to remove the resources definitions altogether.

backend/src/binaryInstaller.ts
```diff
@@ -40,76 +41,56 @@ class BinaryInstaller {
 
   async installBinary(installer: Installer) {
     return new Promise(async (resolve, reject) => {
-      var tmpDir = tmp.dirSync({ unsafeCleanup: true, keep: true })
-
-      // Stop and remove old binaries
-      await this.migrateBinaries(installer.binaryPath())
```
JamieRuderman, line 46: Please add some cleanup of the old locations:
- Uninstall the remoteit agent
- Delete the old binaries

backend/src/binaryInstaller.ts
```diff
@@ -40,76 +41,56 @@ class BinaryInstaller {
...
-      // Download and install binaries
-      await this.download(installer, tmpDir)
-
       const commands = new Command({ onError: reject, admin: true })
 
       if (environment.isWindows) {
         if (!existsSync(environment.binPath)) commands.push(`md "${environment.binPath}"`)
-        commands.push(`move /y "${installer.tempFile}" "${installer.binaryPath()}"`)
-        commands.push(`icacls "${installer.binaryPath()}" /T /C /Q /grant "*S-1-5-32-545:RX"`) // Grant all group "Users" read and execute permissions
+        commands.push(`icacls remoteit /T /C /Q /grant "*S-1-5-32-545:RX"`)
```
JamieRuderman, line 48: Not sure you need this command... if you do it will probably need the .exe
```suggestion
        commands.push(`icacls remoteit.exe /T /C /Q /grant "*S-1-5-32-545:RX"`)
```

backend/src/binaryInstaller.ts
```diff
@@ -40,76 +41,56 @@ class BinaryInstaller {
...
+
+        if (existsSync(`"${environment.mklinkPath}/${installer.binaryName}"`))
```
JamieRuderman, line 50: Please try to make PATH work for Windows.

backend/src/binaryInstaller.ts
```diff
@@ -40,76 +41,56 @@ class BinaryInstaller {
...
-    files.forEach(file => {
-      // Too small to be the desktop app -> must be cli
-      if (existsSync(file) && lstatSync(file).size < 30000000) {
-        Logger.info('MIGRATING DEPRECATED BINARY', { file })
-        commands.push(`"${file}" ${strings.serviceUninstall()}`)
-        commands.push(`"${file}" ${strings.toolsUninstall()}`)
-        toDelete.push(file)
-      } else {
-        Logger.info('DEPRECATED BINARY DOES NOT EXIST', { file })
+  execCommand(command: string, shell: string) {
```
JamieRuderman, line 78: Please remove - unused

backend/src/binaryInstaller.ts
```diff
@@ -124,7 +105,7 @@ class BinaryInstaller {
       const options = { disableGlob: true }
 
       try {
-        rimraf.sync(installer.binaryPath(), options)
+        rimraf.sync(installer.binaryPathCLI(), options)
```
JamieRuderman, line 108: This should only remove the symlinks or the PATH variables in Windows.

backend/src/constants.ts
```diff
@@ -53,24 +58,31 @@ export const SSL_PORT = WEB_PORT - 1
...
+      ? path.resolve('./bin/CLI/x64')
+      : os.platform() === 'win32'
+      ? path.resolve('./resources/x86')
+      : path.resolve('./resources/x64'),
   WIN_ADMIN_SETTINGS: path.resolve('C:/ProgramData/remoteit'),
   WIN_DEPRECATED_BINARIES: [
     path.resolve('C:/Program Files/remoteit/remoteit.exe'),
     path.resolve('/Windows/remoteit.exe'),
   ],
+  WIN_MKLINK_PATH: 'C:/Windows'
```
JamieRuderman, line 87: Can remote since using PATH on windows.

backend/src/environment.ts
```diff
@@ -23,6 +23,7 @@ export class Environment {
   adminUsername: string = ''
   userPath: string
   adminPath: string
+  mklinkPath: string
```
JamieRuderman, line 26: Can remove in this file too

JamieRuderman (changes requested):

JamieRuderman (approved):

PR #300, opened 2020-11-06
DESK-835 ensure signin

JamieRuderman:
### Description

Check signin before all commands
Error messaging cleanup
Fix device sorting with merged devices

### Test plan

1. Uninstall command line tools
2. Restart and Sign in
3. Install the service
4. Go to a device and try to connect
5. You should connect

### Ticket

<https://remoteit.atlassian.net/browse/DESK-835>

### DIFF
--- a/frontend/src/services/graphQL.ts
+++ b/frontend/src/services/graphQL.ts
@@ -27,14 +27,15 @@ export async function graphQLGetErrors(gqlData: any) {
 }
 
 export async function graphQLHandleError(error: AxiosError) {
-  const { auth, backend } = store.dispatch
+  const { auth, backend, ui } = store.dispatch
   console.error('GraphQL fetch error:', error, error?.response?.status)
   if (error?.response?.status === 401) {
-    const { auth } = store.dispatch
     auth.signInError('Session Expired')
   } else if (error?.response?.status === 403) {
     auth.checkSession()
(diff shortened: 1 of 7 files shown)


frontend/src/services/graphQL.ts
```diff
@@ -27,14 +27,15 @@ export async function graphQLGetErrors(gqlData: any) {
...
-  const { auth, backend } = store.dispatch
+  const { auth, backend, ui } = store.dispatch
   console.error('GraphQL fetch error:', error, error?.response?.status)
   if (error?.response?.status === 401) {
-    const { auth } = store.dispatch
     auth.signInError('Session Expired')
   } else if (error?.response?.status === 403) {
     auth.checkSession()
     backend.set({ globalError: error.message })
+  } else if (error.message === 'Network Error') {
```
mauricio0000, line 37: why do you need this else if ?

mauricio0000 (changes requested):

frontend/src/services/graphQL.ts
JamieRuderman: Just hiding the generic "Network Error" messages that appear when waking from sleep. The don't seem to have a response code. Just the message.

mauricio0000 (approved):

PR #301, opened 2020-11-17
DESK-437 Setting launch and copy info as default target attributes

JamieRuderman:
### Description

Setting launch and copy info as default target attributes

### Test plan

1. Sign in
2. Go to your device
3. Add a new service
4. You should see a launch template and command template that you can edit
5. After editing you should see those values as defaults for when you go to connect

### Ticket

<https://remoteit.atlassian.net/browse/DESK-437>


mauricio0000:
@JamieRuderman when a user edit other attributes of the service and don't touch the `Launch URL Template` or `Copy  Command Template` should not update the `Launch URL` or `Copy Command`

mauricio0000 (approved):

PR #302, opened 2020-11-18
[DESK-762] share user fix state check new services

mauricio0000:
### Description

Sharing a second user in the device incorrectly detecting shared-to state

### Test plan

1. Sign in
2. Go to a device
3. Click on share device
4. Select a user
5. Select a second user
4. You should see the combination of services and scripting 

### Ticket

https://remoteit.atlassian.net/browse/DESK-762


JamieRuderman (changes requested):
Looks really good. Only question is if we could show an "indeterminate" state if two users have different permissions for the same checkbox.

image

Let me know if you want to discuss. _Thanks!_

mauricio0000:
https://github.com/remoteit/desktop/pull/302#pullrequestreview-533795350 I am giving priority to the value `true`. what do you think @JamieRuderman ?

JamieRuderman:
@mauricio0000 Let's discuss. Maybe we can't do the indeterminate state.

JamieRuderman:
@mauricio0000 Let's discuss. Maybe we can't do the indeterminate state.

JamieRuderman (changes requested):
@mauricio0000 All looks really great. could you just check that the indeterminate state applies also to the "Allow script execution" checkbox? Code all looked good too. Thanks!

mauricio0000:
@JamieRuderman thank you for the review, I just add indeterminate for scripting

JamieRuderman (changes requested):
@mauricio0000  It all looks good! Working great. 
My only request is to change the component interfaces to replace the "set" commands with "on" commands. Let's discuss! Thanks!

JamieRuderman (approved):
Looks good thanks!

PR #324, opened 2020-12-04
DESK-876 DESK-877 Windows install script update

JamieRuderman:
### Description

Update install script to match internal install/uninstall commands. 
Also hopefully this will fix the connections migration.

### Ticket

<https://remoteit.atlassian.net/browse/DESK-877>
<https://remoteit.atlassian.net/browse/DESK-876>

### DIFF
--- a/backend/src/ConnectionPool.ts
+++ b/backend/src/ConnectionPool.ts
@@ -202,10 +202,9 @@ export default class ConnectionPool {
     const legacyFile = new JSONFile<IConnection[]>(path.join(environment.userPath, 'connections.json'))
     if (legacyFile.exists) {
       Logger.info('MIGRATING LEGACY CONNECTIONS FILE')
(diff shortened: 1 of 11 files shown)


mauricio0000:
@JamieRuderman why do you need `devices.fetch(undefined)`, could you use an optional parameter?

JamieRuderman:
> @JamieRuderman why do you need `devices.fetch(undefined)`, could you use an optional parameter?

@mauricio0000 I really tried hard to get that to work. But since the effect has an other, non optional param after it I couldn't get it to work. I was very frustrating indeed. If you can get it to work please have a look.

mauricio0000 (approved):

JamieRuderman:
@mauricio0000 Okay I just cast the fetch call to 'any' in the backend reference with the optional param. idk why it's not working right, but at least we don't have undefined all over.

PR #345, opened 2020-12-19
DESK-765 connection notifications

JamieRuderman:
### Description

Subscribe to push notifications for device online/offline and connect/disconnect events
Display system notification and update local state

### Test plan

1. Sign in
2. Go to a service and connect
3. You should see the state update and get a connection notice
4. Similarly you should test devices going on/offline and remote connections

### Ticket

<https://remoteit.atlassian.net/browse/DESK-765>


backend/src/ConnectionPool.ts
```diff
@@ -86,9 +86,15 @@ export default class ConnectionPool {
   set = (connection: IConnection, setCLI?: boolean) => {
     if (!connection) Logger.warn('No connections to set!', { connection })
     let instance = this.find(connection.id)
-    if (instance) instance.set(connection, setCLI)
-    else instance = this.add(connection)
-    this.updated()
+    if (instance) {
+      if (JSON.stringify(connection) !== JSON.stringify(instance.params)) {
```
mauricio0000, line 90: could you validate that instance is a object?

backend/src/sharedCopy/constants.ts
```diff
@@ -17,9 +17,11 @@ export const CALLBACK_URL =
   process.env.REACT_APP_CALLBACK_URL || process.env.CALLBACK_URL || process.env.NODE_ENV === 'development'
     ? 'https://dev-auth.internal.remote.it/v1/callback'
     : 'https://auth.api.remote.it/v1/callback'
+export const WEBSOCKET_URL = 'wss://ws.remote.it/beta'
```
mauricio0000, line 20: don't need any environtmen variable?

frontend/src/models/auth.ts
```diff
@@ -100,9 +101,10 @@ export default createModel<RootModel>()({
         dispatch.auth.fetchUser()
       }
     },
-    async authenticated(_: void, rootState: any) {
+    async authenticated(_: void, rootState) {
       if (rootState.auth.authenticated) {
         dispatch.auth.setBackendAuthenticated(true)
+        await cloudController.init()
```
mauricio0000, line 107: could you make any promise all with the await? I mean 4 await in the same method is not good

frontend/src/models/devices.ts
```diff
@@ -243,15 +242,17 @@ export function selectService(state: ApplicationState, serviceId: string) {
...
+  let service: IService | undefined
+  const device = devices.find(
+    d =>
+      d.id === id ||
+      d?.services?.find(s => {
+        if (s.id === id) {
+          service = s
+          return true
+        }
+        return false
```
mauricio0000, line 254: why you need this `return false`?

backend/src/ConnectionPool.ts
JamieRuderman: In this case it doesn't really matter if it's an object or not - just if we've found it in the collection.

backend/src/sharedCopy/constants.ts
JamieRuderman: Not yet anyway - we can add one if needed - we'll probably also need a production url as well at some point.

frontend/src/models/devices.ts
JamieRuderman: Probably dont' need - I'll remove.

frontend/src/models/auth.ts
JamieRuderman: I know some of those require they have data before the others. But I don't need the await in the case of the cloudController init, so I'll remove it.

PR #357, opened 2021-01-08
[DESK-918] airbrake version

mauricio0000:
### Description

add params version to airbrake

### Ticket

https://remoteit.atlassian.net/browse/DESK-918


JamieRuderman (approved):
Thank you!

PR #361, opened 2021-01-12
 DESK-867 filter platform

JamieRuderman:
### Description

Filter devices by Platform

### Ticket

https://remoteit.atlassian.net/browse/DESK-938

### DIFF
--- a/common/types.d.ts
+++ b/common/types.d.ts
@@ -357,6 +357,7 @@ declare global {
     sort?: string
     name?: string
     ids?: string[]
+    platform?: number
   }
 
   interface IRegistration {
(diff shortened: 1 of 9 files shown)


JamieRuderman:
@mauricio0000 Working great - but noticed that the "All" option doesn't select. Could we fix that please?

Also I got this error when clicking on sort by "State".
image

mauricio0000:
@JamieRuderman could you test again?

JamieRuderman:
@mauricio0000 the sorting error is fixed and it defaults to **"all"** checked but when I click **"all"** it doesn't highlight.
image
It should be blue with the check mark.

JamieRuderman:
Approved! @mauricio0000 You are the reviewer - so have to do the actual approval and merge.

mauricio0000 (approved):

PR #362, opened 2021-01-12
Removing maxWidth

GeekCornerGH:
### Description
I have some issues on my computer, and it's awful to have remote.it running in "full screen" while it only occupies part of my screen.

### Checklist

### Test plan

1. Start Remote.it
2. Press Full Screen Button
3. You should have the app in full screen

### Ticket

Nothing 🙂

### DIFF
--- a/src/ElectronApp.ts
+++ b/src/ElectronApp.ts
@@ -113,7 +113,6 @@ export default class ElectronApp {
     this.window = new electron.BrowserWindow({
       width: 800,
       height: 700,
-      maxWidth: 1000,
       minWidth: 525,
       minHeight: 325,
       icon: path.join(__dirname, 'images/icon-64x64.png'),


JamieRuderman:
Hi @GeekCornerGH , thanks for your PR. Unfortunately changing the maxWidth on the electron app window will only display grey side bars:
image
Probably not what you're looking for. However we are going to be working this month on a large screen layout, so we agree that the app needs to get bigger. This is a rough preview: https://www.sketch.com/s/7e847b4f-2010-4e26-8e5f-c9f215ecdc3a/a/3O5qW7w/play

Thanks!

GeekCornerGH:
> Hi @GeekCornerGH , thanks for your PR. Unfortunately changing the maxWidth on the electron app window will only display grey side bars:
> image
> Thanks!

This would be better than a smaller Window (devtools for example)

JamieRuderman:
@GeekCornerGH It seems strange to allow it to be larger just to show blank columns. If you really want to see it larger you could view it in your browser at http://localhost:29999. For the dev tools, you can just pop them out into their own window. Hopefully these options will work for you while we implement the large screen layouts.

JamieRuderman:
@GeekCornerGH So I've figured out how to make it go full screen in windows without removing the floating window size constraints. You can see what I did here: https://github.com/remoteit/desktop/pull/368

I'm going to close this PR. Thanks for your feedback.

PR #363, opened 2021-01-12
[DESK-910] show license notice just for device owner

mauricio0000:
### Description

show license notice just for the device owner

### Test plan

1. Sign in
2. Go to the device list
3. Click in one AWS device of yourself
4. You should see a message if your plan is trial

### Ticket

https://remoteit.atlassian.net/browse/DESK-910

### DIFF
--- a/frontend/src/components/LicensingNotice.tsx
+++ b/frontend/src/components/LicensingNotice.tsx
@@ -16,9 +16,11 @@ const learnMoreLink = (
 )
 
 export const LicensingNotice: React.FC<Props> = props => {
-  const { noticeType, license, serviceLimit, upgradeUrl = '' } = useSelector((state: ApplicationState) =>
-    selectLicense(state, props.device ? lookupLicenseProductId(props.device) : props.license?.plan.product.id)
-  )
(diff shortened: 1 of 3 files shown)


JamieRuderman (changes requested):
Hmmm... I think I pointed you in the wrong direction, sorry. Looking more I think this is what we want:

```
  const { noticeType, license, serviceLimit, upgradeUrl = '' } = useSelector((state: ApplicationState) => {
    let productId = props.license?.plan.product.id
    if (props.device && state.auth.user?.id === props.device.owner.id) productId = lookupLicenseProductId(props.device)
    return selectLicense(state, productId)
  })
```

Added to `LicensingNotice.tsx` and `LicensingServiceNotice.tsx`

Let me know if that works better. Thanks!

JamieRuderman:
@mauricio0000 Also I've noticed that the `ApplicationState` seems to have lost it's TypeScript info in `useSelector((state: ApplicationState) ` Do you have any idea of how to fix that?

Thanks!

JamieRuderman (approved):
I think this is ok for now, but would probably be good to figure out how to move this duplicated logic into the selectLicense function

PR #373, opened 2021-01-20
[DESK-794] show linked user in access device

mauricio0000:
### Description

show linked user in access device

### Test plan

1. Sign in
2. Go to device list
3. Click on one of your own devices
4. Click in Shared Users
4. You should see a list of `DEVICE LIST SHARED`

### Ticket

https://remoteit.atlassian.net/browse/DESK-794


JamieRuderman (changes requested):
Let's not remove the user from the standard sharing list if they are also shared with device list sharing. They should appear in both, so that a user can remove the standard share if they'd like.

mauricio0000:
@JamieRuderman updated

JamieRuderman (approved):
thank you!

PR #397, opened 2021-02-02
DESK-930 Announcements

JamieRuderman:
### Description

Provide in app messaging

### Test plan

1. Sign in
2. See announcements indicator
3. Click it and see announcements
4. You should see the indicator disappear after reading.

### Ticket

<https://remoteit.atlassian.net/browse/DESK-930>


frontend/src/models/announcements.ts
```diff
@@ -0,0 +1,119 @@
...
+      return state
+    },
+  },
+})
+
+export function selectAnnouncements(state: ApplicationState, unread?: boolean) {
+  return state.announcements.all.filter(a => !unread || !a.read)
+}
+
+const TEST_DATA: any = [
```
mauricio0000, line 81: do you need this TEST_DATA?

frontend/src/pages/AnnouncementsPage.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { AnnouncementCard } from '../components/AnnouncementCard'
+import { Container } from '../components/Container'
+import { Title } from '../components/Title'
+import { Icon } from '../components/Icon'
+
+export const AnnouncementsPage = () => {
+  const enabled = useSelector((state: ApplicationState) => selectAnnouncements(state))
+
+  return (
+    <Container
```
mauricio0000, line 15: are you to show this component if `enabled` doesn't have data?

frontend/src/models/announcements.ts
JamieRuderman: Don't need, but it's useful for testing.

frontend/src/pages/AnnouncementsPage.tsx
JamieRuderman: Yes we would show the page anyway. It just would be empty until the announcements are loaded. I rename enabled to annoucements. It used to be that I was checking enabled, but Benoit said he's checking that cloud side so I removed it.

PR #409, opened 2021-02-12
[DESK-957] ignore read only error

mauricio0000:
### Description

[DESK-957] ignore read-only file system error

### Ticket

https://remoteit.atlassian.net/browse/DESK-957

[DESK-957]: https://remoteit.atlassian.net/browse/DESK-957


JamieRuderman (approved):
Nice work. Thanks!

PR #426, opened 2021-03-08
[DESK-926] bind ip edit and validation

mauricio0000:
### Description

bind ip edit and validation

### Test plan
https://remoteit.atlassian.net/wiki/x/AYBHO

https://remoteit.atlassian.net/browse/DESK-926

### DIFF
--- a/backend/src/sharedCopy/constants.ts
+++ b/backend/src/sharedCopy/constants.ts
@@ -36,7 +36,7 @@ export const REGEX_PORT_SAFE = /[^0-9]+/g
 export const REGEX_NUMERIC_VALUE = /=(\d+)/
 export const REGEX_LAST_NUMBER = /-*\d*$/
 export const REGEX_VALID_IP = /^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]):[0-9]+$/
-export const REGEX_VALID_HOSTNAME = /^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9]):[0-9]+$/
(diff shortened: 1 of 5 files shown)


JamieRuderman (changes requested):
I checked out the branch, but couldn't get to the page, so made some fixes and committed them.  Please pull. Also the Save button doesn't enable so you can't save, except if you change the bind IP address - then if you have an error it shows the error and then saves anyway. Also I don't think we need the IP address validation because I think you might want to enter in a dns name like "localhost" or "system.local"

JamieRuderman (changes requested):
I tested and when I turned it on and clicked save it didn't turn on and save.

backend/src/sharedCopy/constants.ts
```diff
@@ -36,7 +36,7 @@ export const REGEX_PORT_SAFE = /[^0-9]+/g
 export const REGEX_NUMERIC_VALUE = /=(\d+)/
 export const REGEX_LAST_NUMBER = /-*\d*$/
 export const REGEX_VALID_IP = /^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5]):[0-9]+$/
-export const REGEX_VALID_HOSTNAME = /^(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\-]*[a-zA-Z0-9])\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\-]*[A-Za-z0-9]):[0-9]+$/
```
mauricio0000, line 39: this REGEX was checking port

frontend/src/pages/LanSharePage/LanSharePage.tsx
```diff
@@ -56,30 +56,61 @@ export const LanSharePage: React.FC = () => {
   const history = useHistory()
   const css = useStyles()
   const [currentHost, setCurrentHost] = useState((connection && connection.host) || IP_PRIVATE)
-
-  if (!connection || !service) return null
+  const [error, setError] = useState<string>()
+  const [bindIP, setBindIP] = useState(IP_OPEN)
```
JamieRuderman, line 60:
```suggestion
```
I don't think you want a new state for `bindIP`. `bindIP` is the same as `currentHost` I think.

JamieRuderman (changes requested):

JamieRuderman (approved):

PR #432, opened 2021-03-15
[DESK-987] shutdown all agents click in restart

mauricio0000:
### Description

shupdown all agents click in restart

### Test plan

1. Sign in on old version 
2. You should see modal to upgrade version
3. Click restart
4. You should see that every service shutdown

### Ticket

https://remoteit.atlassian.net/browse/DESK-987

### DIFF
--- a/backend/src/Controller.ts
+++ b/backend/src/Controller.ts
@@ -182,8 +182,9 @@ class Controller {
     app.quit()
   }
 
-  restart = (update?: string) => {
+  restart = async (update?: string) => {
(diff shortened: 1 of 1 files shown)


JamieRuderman (changes requested):
@mauricio0000 we shouldn't stop the connections as you're doing here - we need to stop the system agent. That is done by running CLI.ts -> serviceUninstall()

Sorry for the confusion!

JamieRuderman (approved):
perfect!

PR #443, opened 2021-03-23
DESK-997 Changing restore state from model to route

JamieRuderman:
### Test plan

1. Sign in on a system without a device
2. Click restore button
3. Click the "Register services" button instead of restoring
4. You should see the setup screen and then after finished should see the device list as normal, not in the restore state.

### Ticket

<https://remoteit.atlassian.net/browse/DESK-997 >

### DIFF
--- a/frontend/src/components/DeviceSetupItem/DeviceSetupItem.tsx
+++ b/frontend/src/components/DeviceSetupItem/DeviceSetupItem.tsx
@@ -69,14 +69,7 @@ export const DeviceSetupItem: React.FC<{ restore?: boolean }> = ({ restore }) =>
               <Link onClick={() => history.push('/devices')}>cancel</Link>
             </Typography>
           ) : (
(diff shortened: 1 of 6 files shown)


frontend/src/components/DeviceSetupItem/DeviceSetupItem.tsx
```diff
@@ -68,15 +66,15 @@ export const DeviceSetupItem: React.FC = () => {
...
+              <Link onClick={() => history.push('/devices')}>cancel</Link>
             </Typography>
           ) : (
             <Button
               variant="outlined"
               size="small"
               onClick={() => {
-                ui.set({ restore: true })
-                history.push('/devices')
+                console.log('---------click------------')
```
mauricio0000, line 76: could we remove it?

mauricio0000 (approved):

PR #449, opened 2021-03-26
[DESK-1013] check update version beta alpha

mauricio0000:
### Description

check-update version also using beta  and alpha

`semverCompare` avoid check version if the current version is alpha o beta

before to compare, this change is sending just version without `alpha` or `beta`

### Test plan

1. Use a version alpha or beta newest than the last stable release
2. You should not see a modal to update

### Ticket

https://remoteit.atlassian.net/browse/DESK-1038

### DIFF
--- a/frontend/src/pages/SettingsPage/SettingsPage.tsx
+++ b/frontend/src/pages/SettingsPage/SettingsPage.tsx
@@ -23,7 +23,7 @@ import { Logo } from '../../components/Logo'
 import analyticsHelper from '../../helpers/analyticsHelper'
 
 export const SettingsPage: React.FC<{ singlePanel?: boolean }> = ({ singlePanel }) => {
-  const { showReports, os, installing, cliVersion, preferences, targetDevice, notOwner, remoteUI } = useSelector(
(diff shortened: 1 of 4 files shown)


JamieRuderman (changes requested):
Let's not change the semverCompare input. We do want alpha or beta to update. The issue in the ticket was about it trying to downgrade from a later version. If you can't reproduce please let me know. You might also want to add a unit test for the linux updating.

frontend/src/pages/SettingsPage/SettingsPage.tsx
```diff
@@ -21,9 +21,10 @@ import { spacing } from '../../styling'
 import { Title } from '../../components/Title'
 import { Logo } from '../../components/Logo'
 import analyticsHelper from '../../helpers/analyticsHelper'
+import { AutoUpdateSetting } from '../../components/AutoUpdateSetting/AutoUpdateSetting'
```
JamieRuderman, line 24:
```suggestion
import { AutoUpdateSetting } from '../../components/AutoUpdateSetting'
```

JamieRuderman (approved):
Looks good thank you!

PR #453, opened 2021-03-30
DESK-1056 duration end time

JamieRuderman:
### Description

Updating connection duration display with end time

### Ticket

<https://remoteit.atlassian.net/browse/DESK-1056>

### DIFF
--- a/frontend/src/components/Duration/Duration.tsx
+++ b/frontend/src/components/Duration/Duration.tsx
@@ -2,14 +2,23 @@ import React, { useState } from 'react'
 import { useInterval } from '../../hooks/useInterval'
 import humanize from 'humanize-duration'
 
-export const dateOptions = { weekday: 'long', year: 'numeric', month: 'long', day: 'numeric' }
+export const dateOptions: Intl.DateTimeFormatOptions = {
+  weekday: 'long',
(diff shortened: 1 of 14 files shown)


frontend/src/components/Duration/Duration.tsx
```diff
@@ -2,14 +2,23 @@ import React, { useState } from 'react'
...
-  const [now, setNow] = useState(Date.now())
+export const Duration: React.FC<{ startTime?: number; endTime?: number; ago?: boolean }> = ({
+  startTime,
+  endTime = Date.now(),
+  ago = false,
+}) => {
+  const [now, setNow] = useState<number>(endTime)
   const aDay = 1000 * 60 * 60 * 24
 
   useInterval(() => {
```
mauricio0000, line 20: I think that it should be a set timeout not setInterval, maybe double-check the hooks because we want to setup. 

I am not 100% sure

JamieRuderman: Yea interval is correct for this because it's repeating forever will be a little lighter on resources. And the useInterval will clean up when the component is teared down

mauricio0000 (approved):

PR #470, opened 2021-04-15
[DESK-1060] clear all connection

mauricio0000:
### Description

[DESK-1060] clear all connection

### Ticket

https://remoteit.atlassian.net/browse/DESK-1060

[DESK-1060]: https://remoteit.atlassian.net/browse/DESK-1060

### DIFF
--- a/backend/src/CLI.ts
+++ b/backend/src/CLI.ts
@@ -264,6 +264,10 @@ export default class CLI {
     await this.exec({ cmds: [strings.serviceUninstall()], admin: true })
   }
 
+  async restart() {
+    await this.exec({ cmds: [strings.serviceRestart()], admin: true})
+  }
+
   async signIn() {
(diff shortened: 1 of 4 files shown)


JamieRuderman (changes requested):
@mauricio0000 This isn't what we need - I looked into the code some more and I think this is what should be done:

In the ConnectionPool.ts file `constructor()` add an event took into the binary installer installed event:
```
EventBus.on(Binary.EVENTS.installed, this.clearErrors)
```

Then the `clearErrors` method say something like:
```
  clearErrors = () => {
    this.pool.forEach(connection => (connection.params.error = undefined))
    this.updated()
  }
```
This is untested, but should be all that's needed. Thanks!

JamieRuderman (approved):
thanks!

PR #472, opened 2021-04-17
[DESK-1076] sort services

mauricio0000:
### Description

 sort services

### Test plan

1. Sign in
2. Go to device
3. Click the new icon to sort services
4. You should see sorted services as you selected 

### Ticket
https://remoteit.atlassian.net/browse/DESK-1076

### DIFF
--- a/frontend/src/components/SortServices.tsx
+++ b/frontend/src/components/SortServices.tsx
@@ -1,44 +1,62 @@
 import React, { useEffect } from 'react'
 import { Badge, IconButton, Menu, MenuItem } from '@material-ui/core'
 import { Icon } from './Icon'
-import { useDispatch } from 'react-redux'
-import { Dispatch } from '../store'
+import { useDispatch, useSelector } from 'react-redux'
+import { ApplicationState, Dispatch } from '../store'
 import { makeStyles } from '@material-ui/core/styles'
 import { fontSize } from '@remote.it/components/lib/styles/variables'
 
 export interface ISortService {
-  value: number
   name: string
   sortService: (a: IService, b: IService) => number
+  icon: string
 }
-const options: ISortService[] = [
-  {
-    value: 0,
+
+export interface IOptionServiceSort {
+  ATOZ: ISortService
+  ZTOA: ISortService
+  NEWEST: ISortService
+  OLDEST: ISortService
+}
+
+export const optionSortServices: IOptionServiceSort = {
+  ATOZ: {
     name: 'Alpha A-Z',
     sortService: (a: IService, b: IService) =>
       a.name.toLowerCase() > b.name.toLowerCase() ? 1 : a.name.toLowerCase() < b.name.toLowerCase() ? -1 : 0,
+    icon: 'up',
   },
-  {
-    value: 1,
+  ZTOA: {
     name: 'Alpha Z-A',
     sortService: (a: IService, b: IService) =>
       a.name.toLowerCase() < b.name.toLowerCase() ? 1 : a.name.toLowerCase() > b.name.toLowerCase() ? -1 : 0,
+    icon: 'down',
   },
-  { value: 2, name: 'Newest first', sortService: (a: IService, b: IService) => (a.createdAt > b.createdAt ? 1 : -1) },
-  { value: 3, name: 'Oldest first', sortService: (a: IService, b: IService) => (a.createdAt < b.createdAt ? 1 : -1) },
-]
+  NEWEST: {
+    name: 'Newest first',
+    sortService: (a: IService, b: IService) => (a.createdAt > b.createdAt ? 1 : -1),
+    icon: 'up',
+  },
+  OLDEST: {
+    name: 'Oldest first',
+    sortService: (a: IService, b: IService) => (a.createdAt < b.createdAt ? 1 : -1),
+    icon: 'down',
+  },
+}
 
 export const SortServices: React.FC = () => {
   const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null)
   const { devices } = useDispatch<Dispatch>()
+  const { sortService } = useSelector((state: ApplicationState) => ({
+    sortService: state.devices.sortServiceOption,
+  }))
+  const option = optionSortServices[`${sortService}`]
+
   const open = Boolean(anchorEl)
   const css = useStyles()
-  const [icon, setIcon] = React.useState('sort-amount-down')
-  const [dot, setDot] = React.useState(false)
-  const [optionSelected, setOptionSelected] = React.useState<ISortService>()
 
   useEffect(() => {
-    servicesSort(options[0])
+    servicesSort(sortService)
   }, [])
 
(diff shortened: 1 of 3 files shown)


JamieRuderman (changes requested):
Please make sure that the menu isn't cut off. Also use the small size for the menu:
image

Here please move the icon to the right of the column. Also change the direction of the arrow on the icon based on the sort direction, and add a blue dot on it if it's been changed from default. Please see the device list filter for reference. 
image

Thank you!

JamieRuderman (changes requested):
- Move icon to right of column
- Remove change indicator if it's returned to default state
- The icon isn't constantly pointing the same direction when sorted - I can change direction and it doesn't change, or if I re-select the same it changes.

frontend/src/components/SortServices.tsx
```diff
@@ -0,0 +1,98 @@
...
+        sortService = (a: IService, b: IService) =>
+          a.name.toLowerCase() < b.name.toLowerCase() ? 1 : a.name.toLowerCase() > b.name.toLowerCase() ? -1 : 0
+        break
+      case 2:
+        sortService = (a: IService, b: IService) => (a.createdAt > b.createdAt ? 1 : -1)
+        break
+      case 3:
+        sortService = (a: IService, b: IService) => (a.createdAt < b.createdAt ? 1 : -1)
+        break
+    }
```
JamieRuderman, line 55: What about adding these sort functions to the options on line 9 above, so that you can then save the option index to the device model. That way you can read the sort function AND the selected option out of the model at the same time so that you can restore the icon state when leaving and returning to the page.

JamieRuderman (changes requested):
Looking great - just one small change to allow restoring of icon state

frontend/src/components/SortServices.tsx
```diff
@@ -0,0 +1,91 @@
...
+import { Dispatch } from '../store'
+import { makeStyles } from '@material-ui/core/styles'
+import { fontSize } from '@remote.it/components/lib/styles/variables'
+
+export interface ISortService {
+  value: number
+  name: string
+  sortService: (a: IService, b: IService) => number
+}
+const options: ISortService[] = [
```
JamieRuderman, line 14: Let's export this so it can be referenced by the devicesPage.tsx file.

Also, instead of an array it could be a lookup. For example:
```
{
   ATOZ: { sort: ..., name: ..., icon: 'up' },
   ZTOA: ...
   NEWEST: ...
   OLDEST:...
}
```
That way when you save the selected sort state  you just write the ENUM to the model:
`devices.set({ serviceSort }: { serviceSort: 'ATOZ' | 'ZTOA' | 'NEWEST' | 'OLDEST' })`

frontend/src/components/SortServices.tsx
```diff
@@ -0,0 +1,91 @@
...
+]
+
+export const SortServices: React.FC = () => {
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null)
+  const { devices } = useDispatch<Dispatch>()
+  const open = Boolean(anchorEl)
+  const css = useStyles()
+  const [icon, setIcon] = React.useState('sort-amount-down')
+  const [dot, setDot] = React.useState(false)
+  const [optionSelected, setOptionSelected] = React.useState<ISortService>()
```
JamieRuderman, line 38: We don't want local state, this should be removed, and only use the device model global state so that the setting is recalled after page changes.
```suggestion
```

frontend/src/components/SortServices.tsx
```diff
@@ -0,0 +1,91 @@
...
+  { value: 3, name: 'Oldest first', sortService: (a: IService, b: IService) => (a.createdAt < b.createdAt ? 1 : -1) },
+]
+
+export const SortServices: React.FC = () => {
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null)
+  const { devices } = useDispatch<Dispatch>()
+  const open = Boolean(anchorEl)
+  const css = useStyles()
+  const [icon, setIcon] = React.useState('sort-amount-down')
+  const [dot, setDot] = React.useState(false)
```
JamieRuderman, line 37: Is there a reason to set the dot state instead of calculating it inline?

frontend/src/components/SortServices.tsx
```diff
@@ -0,0 +1,91 @@
...
+    option.value === 0 || option.value === 2 ? setIcon('sort-amount-down') : setIcon('sort-amount-up')
+    setDot(!!option.value)
+    devices.set({ sortServiceOption: option })
+    handleClose()
+  }
+
+  return (
+    <>
+      <IconButton aria-label="more" aria-controls="long-menu" aria-haspopup="true" onClick={handleClick}>
+        {dot ? (
```
JamieRuderman, line 63:
```suggestion
        {!!option.value ? (
```

frontend/src/components/SortServices.tsx
```diff
@@ -0,0 +1,91 @@
...
+  { value: 2, name: 'Newest first', sortService: (a: IService, b: IService) => (a.createdAt > b.createdAt ? 1 : -1) },
+  { value: 3, name: 'Oldest first', sortService: (a: IService, b: IService) => (a.createdAt < b.createdAt ? 1 : -1) },
+]
+
+export const SortServices: React.FC = () => {
+  const [anchorEl, setAnchorEl] = React.useState<null | HTMLElement>(null)
+  const { devices } = useDispatch<Dispatch>()
+  const open = Boolean(anchorEl)
+  const css = useStyles()
+  const [icon, setIcon] = React.useState('sort-amount-down')
```
JamieRuderman, line 36: Probably don't need.
```suggestion
```

frontend/src/components/SortServices.tsx
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  const handleClose = () => {
+    setAnchorEl(null)
+  }
+
+  const servicesSort = async (option: ISortService) => {
+    setOptionSelected(option)
+    option.value === 0 || option.value === 2 ? setIcon('sort-amount-down') : setIcon('sort-amount-up')
+    setDot(!!option.value)
```
JamieRuderman, line 55:
```suggestion
```

frontend/src/components/SortServices.tsx
```diff
@@ -0,0 +1,91 @@
...
+    devices.set({ sortServiceOption: option })
+    handleClose()
+  }
+
+  return (
+    <>
+      <IconButton aria-label="more" aria-controls="long-menu" aria-haspopup="true" onClick={handleClick}>
+        {dot ? (
+          <Badge variant="dot" badgeContent={true} color="primary">
+            <Icon name={icon} size="md" />
```
JamieRuderman, line 65:
```suggestion
            <Icon name={`sort-amount-${option.icon}`} size="md" />
```

JamieRuderman (changes requested):
Please see comments thanks!

JamieRuderman (approved):
Looks good thanks!

PR #476, opened 2021-04-22
[DESK-1074] clear connections when change user signed

mauricio0000:
### Description
clear connections when change user signed

### Test plan

1. Sign in
2. Create  connections
3. Logout
4. You should not see on the menu electron connections

### Ticket

https://remoteit.atlassian.net/browse/DESK-1074

image

### DIFF
--- a/backend/src/User.ts
+++ b/backend/src/User.ts
@@ -6,6 +6,7 @@ import Logger from './Logger'
 import EventBus from './EventBus'
 import path from 'path'
 import { r3 } from './remote.it'
+import ConnectionPool from './ConnectionPool'
 
 const d = debug('r3:backend:User')
 
@@ -80,6 +81,7 @@ export class User {
 
     Logger.info('SIGN OUT USER', { user: this })
 
+    EventBus.emit(ConnectionPool.EVENTS.updated, [])
     EventBus.emit(User.EVENTS.signedOut)
     try {
       rimraf.sync(path.join(environment.userPath, 'user.json'), { disableGlob: true })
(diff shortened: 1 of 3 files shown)


JamieRuderman (approved):
thanks

JamieRuderman (changes requested):
@mauricio0000 I think that we should be clearing the actual data in memory instead of hiding the menu. Also it looks like if you hide this it will also hide the sign in menu items.

Please see if the `signOut` method is called in the backend controller - it should be clearing those connections.

JamieRuderman (changes requested):
This is great, I think you've found the issue. It's not updating the backend state after signing out... but I'm wondering if it would be better to add the update to the `clearMemory` method in the connection pool controller instead of adding it to the user class.

Let's just add `EventBus.emit(ConnectionPool.EVENTS.updated, [])` to the end of the `clearMemory` method please. And verify that it's not erasing the connections so they are still there when you sign back in.

Thanks!

backend/src/User.ts
```diff
@@ -80,6 +81,7 @@ export class User {
 
     Logger.info('SIGN OUT USER', { user: this })
 
+    EventBus.emit(ConnectionPool.EVENTS.updated, [])
```
mauricio0000, line 84: @JamieRuderman do you mean that we should remove this line?

JamieRuderman: Yes remove and add to the `clearMemory` method.

backend/src/User.ts
```diff
@@ -6,6 +6,7 @@ import Logger from './Logger'
 import EventBus from './EventBus'
 import path from 'path'
 import { r3 } from './remote.it'
+import ConnectionPool from './ConnectionPool'
```
JamieRuderman, line 9:
```suggestion
```

JamieRuderman (changes requested):
One small one

JamieRuderman (approved):
Thanks!

PR #477, opened 2021-04-22
DESK-1037 DESK-1085 public connections

JamieRuderman:
### Description

- Fix clearing errors to only happen after install
- Fix certificate inclusion for mac
- Adding support for public connections
- Trying to fix missing h1 font loading
- Fixing typescript issues
- Don't show share button on shared devices
- Allow clearing of connections without loosing state
- Show disabled settings as lighter grey
- DESK-1085 LAN sharing not indicating when off
- Fixing reset connection names generation

### Test plan

Testing proxy only connections by enabling test ui then:

1. Click new connection and select a web service
2. Click the shareable link toggle at the bottom
3. Click start connection
4. You should see the public proxy connection details

### Ticket

<https://remoteit.atlassian.net/browse/DESK-1037>
<https://remoteit.atlassian.net/browse/DESK-1085>

### DIFF
--- a/frontend/src/components/PublicSetting.tsx
+++ b/frontend/src/components/PublicSetting.tsx
@@ -1,0 +1,54 @@
+import React from 'react'
+import { Collapse } from '@material-ui/core'
+import { setConnection } from '../helpers/connectionHelper'
+import { IP_OPEN, IP_LATCH } from '../shared/constants'
+import { InlineSelectSetting } from './InlineSelectSetting'
+import { ListItemSetting } from './ListItemSetting'
+
+export const PublicSetting: React.FC<{ service: IService; connection?: IConnection }> = ({ service, connection }) => {
+  if (!connection) return null
+  if (service.attributes.route === 'p2p') return null
+
+  const disabled = connection.enabled
+  const subLabel =
+    connection.publicRestriction === IP_LATCH
+      ? 'The connection will latch onto the first device to connect with IP restriction.'
+      : 'Any device will be able to connect while the connection is active.'
+
+  return (
+    <>
+      <ListItemSetting
+        label="Shareable Link"
+        subLabel={subLabel}
+        disabled={disabled}
+        icon="globe-americas"
+        toggle={!!connection.public}
+        onClick={() =>
+          connection &&
+          setConnection({
+            ...connection,
+            public: !connection.public,
+          })
+        }
+      />
+      <Collapse in={connection.public} timeout={400}>
+        <InlineSelectSetting
+          label="Security"
+          disabled={disabled || !connection.public}
+          value={connection.publicRestriction}
+          values={[
+            { name: 'IP Latching', key: IP_LATCH },
+            { name: 'None', key: IP_OPEN },
+          ]}
+          onSave={key => {
+            connection &&
+              setConnection({
+                ...connection,
+                publicRestriction: key.toString(),
+              })
+          }}
+        />
+      </Collapse>
+    </>
+  )
+}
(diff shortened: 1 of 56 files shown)


frontend/src/components/PublicSetting.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  return (
+    <>
+      <ListItemSetting
+        label="Shareable Link"
+        subLabel={subLabel}
+        disabled={disabled}
+        icon="globe-americas"
+        toggle={!!connection.public}
+        onClick={() =>
```
mauricio0000, line 26: maybe in this case create a cont with the function to avoid re-render

like 

``const handleClick = () => {
  connection &&
          setConnection({
            ...connection,
            public: !connection.public,
          })
}``


`` onClick={handleClick}``

frontend/src/models/connections.ts
```diff
@@ -0,0 +1,160 @@
...
+    async updateConnection(connection: IConnection, globalState) {
+      globalState.connections.all.some((c, index) => {
+        if (c.id === connection.id) {
+          globalState.connections[index] = connection
+          dispatch.backend.set({ connections: globalState.connections.all })
+          if (connection) return true
+        }
+        return false
+      })
+    },
```
mauricio0000, line 30: Maybe something like that? 

```
async updateConnection(connection: IConnection, {connections}: ApplicationState) {
      connections.all.some((c, index) => {
        if (c.id === connection.id) {
          connections[index] = connection
          dispatch.backend.set({ connections: connections.all })
          if (connection) return true
        }
        return false
      })
    },
```

frontend/src/models/connections.ts
```diff
@@ -0,0 +1,160 @@
...
+    async restoreConnections(connections: IConnection[], globalState) {
+      connections.forEach(async connection => {
+        // data missing from cli if our connections file is lost
+        if (!connection.owner) {
+          const [service] = selectById(globalState, connection.id)
+          if (service) {
+            connection = { ...newConnection(service), ...connection }
+            setConnection(connection)
+          } else {
+            console.warn('No service found for connection', connection.id)
```
mauricio0000, line 41: do you need that?

frontend/src/models/connections.ts
```diff
@@ -0,0 +1,160 @@
...
+            // fetch device?
+            const device = await dispatch.devices.fetchSingle({ id: connection.id, hidden: true })
+            console.log('FETCHED DEVICE RETURNED:', device)
```
mauricio0000, line 44: same

frontend/src/models/connections.ts
```diff
@@ -0,0 +1,160 @@
...
+          'Content-Type': 'application/json',
+          developerKey: DEVELOPER_KEY,
+          authorization: token,
+        },
+      }
+    },
+
+    async proxyConnect(connection: IConnection): Promise<any> {
+      const data = { wait: 'true', hostip: connection.publicRestriction, deviceaddress: connection.id }
+      console.log('PROXY CONNECT', `${API_URL}/device/connect`, data)
```
mauricio0000, line 65: dev logs

frontend/src/models/connections.ts
JamieRuderman: Yes - this will help for troubleshooting

frontend/src/components/PublicSetting.tsx
JamieRuderman: I don't believe that having the handler inline as an anonymous function would cause a re-render, is that true?

mauricio0000 (approved):

PR #481, opened 2021-05-04
[DESK-1099] help menu

mauricio0000:
### Description

help menu

### Test plan

image

### Ticket

https://remoteit.atlassian.net/browse/DESK-1099


src/ElectronApp.ts
```diff
@@ -30,6 +30,23 @@ export default class ElectronApp {
...
     this.protocol = process.env.NODE_ENV === 'development' ? DEEP_LINK_PROTOCOL_DEV : DEEP_LINK_PROTOCOL
 
+    const template = [
+      {
+        label: 'Help',
+        submenu: [
+          {
+            label: 'Learn More',
+            click() {
+              electron.shell.openExternal('https://support.remote.it/hc/en-us')
```
JamieRuderman, line 40: Could you check with @bstrech and see if there is a short.io link we should use? Also is there a way to keep the other default menus? They seem to be lost.

JamieRuderman (changes requested):

PR #482, opened 2021-05-06
[DESK-1102] allow change api url

mauricio0000:
### Description

[DESK-1102] allow change api url

### Test plan

1. Sign in
2. Go to...
3. Click...
4. You should see X, Y, Z...

### Ticket

https://remoteit.atlassian.net/browse/DESK-1102

[DESK-1102]: https://remoteit.atlassian.net/browse/DESK-1102


frontend/src/services/graphQL.ts
```diff
@@ -1,15 +1,23 @@
 import axios, { AxiosError, AxiosResponse } from 'axios'
 import { getToken } from '../services/remote.it'
 import { store } from '../store'
-import { version } from '../../package.json'
 import { GRAPHQL_API, GRAPHQL_BETA_API } from '../shared/constants'
+import { version } from '../../package.json'
+
+export function getApiURL(isRest?: boolean): string {
```
JamieRuderman, line 7:
```suggestion
export function getApiURL(): string {
```

Adding restAPI to graphQL doesn't do anything. The restAPI is actually defined in `shared/constants.ts` as `API_URL` so this setting will have to override all uses of `API_URL` in the code base as well.

frontend/src/services/graphQL.ts
```diff
@@ -1,15 +1,23 @@
...
 
-export function getApi(): string {
-  const { backend } = store.getState()
-  const { overrides } = backend.environment
   let beta = version.includes('alpha')
-  if (backend.preferences.switchApi) beta = !beta
-  return beta ? overrides?.betaApiURL || GRAPHQL_BETA_API : overrides?.apiURL || GRAPHQL_API
+  if (isRest) {
+    return apiURL || ''
+  } else {
```
JamieRuderman, line 14: No (isRest) logic should be in graphQL code. See comment above.

JamieRuderman (changes requested):
Great work, just some tweaks.

JamieRuderman (approved):
Looks good thanks!

PR #500, opened 2021-06-30
[DESK-1081] connection port

mauricio0000:
### Description

[DESK-1081] connection port default on reset

### Test plan

1. select your own device
2. select a service
3. edit the default port field
4. connect service
5. check under the connect button the address

### Ticket

https://remoteit.atlassian.net/browse/DESK-1081

[DESK-1081]: https://remoteit.atlassian.net/browse/DESK-1081


JamieRuderman (commented):
Although this will fix the issue won't it break the connect button so that it won't send the connection info that you've configured? I'm not sure what Gary's actual problem is in the ticket. The feature seems to be working correctly. Once a port is set it's not supposed to change. So it might not actually be a bug. Not sure.

PR #511, opened 2021-07-12
MOB-550 On boarding guide fixes

JamieRuderman:
### Description

- Adding explicit active state and saving to local storage for guides
- Adding tracking header to websocket authentication MOB-550

### DIFF
--- a/frontend/src/services/Controller.ts
+++ b/frontend/src/services/Controller.ts
@@ -116,6 +116,7 @@ function getEventHandlers() {
       await store.dispatch.licensing.fetch()
       await store.dispatch.accounts.init()
       await store.dispatch.devices.init()
+      await store.dispatch.ui.init()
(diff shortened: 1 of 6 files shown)


frontend/src/services/Controller.ts
```diff
@@ -116,6 +116,7 @@ function getEventHandlers() {
       await store.dispatch.licensing.fetch()
       await store.dispatch.accounts.init()
       await store.dispatch.devices.init()
+      await store.dispatch.ui.init()
```
mauricio0000, line 119: @JamieRuderman could we use an await promise all ?

JamieRuderman: @mauricio0000 if we did Promise.all it would resolve all the promises asynchronously... I think it's safer to do them synchronously as above.

mauricio0000 (approved):

PR #540, opened 2021-08-30
[DESK-1193] remove r3 rest api token and use cli

mauricio0000:
### Description

[DESK-1193] remove r3 rest api token and use cli

### Test plan

1. Sign in
2. Go to app
3. You should see working desktop

### Ticket

https://remoteit.atlassian.net/browse/DESK-1193

[DESK-1193]: https://remoteit.atlassian.net/browse/DESK-1193

### DIFF
--- a/backend/src/User.ts
+++ b/backend/src/User.ts
@@ -5,7 +5,6 @@ import debug from 'debug'
 import Logger from './Logger'
 import EventBus from './EventBus'
 import path from 'path'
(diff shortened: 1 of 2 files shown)


JamieRuderman (changes requested):
@mauricio0000 there still has to be validation of the credentials with the cloud API. You will have to do the same axios call to authhash login to validate that the credentials are correct.

PR #542, opened 2021-09-01
[DESK-1204] headless no autoupdate

mauricio0000:
### Description

[DESK-1204] headless no autoupdate

### Test plan

tar -xf remoteit-headless.tgz

node package/build

checked update should be avoid

### Ticket

https://remoteit.atlassian.net/browse/DESK-1204

[DESK-1204]: https://remoteit.atlassian.net/browse/DESK-1204

### DIFF
--- a/src/ElectronApp.ts
+++ b/src/ElectronApp.ts
@@ -49,7 +49,9 @@ export default class ElectronApp {
     EventBus.on(EVENTS.open, this.openWindow)
   }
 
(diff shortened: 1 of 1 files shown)


JamieRuderman (changes requested):
@mauricio0000 I think it would be better to have the isHeadless validation done when it's checking for updates instead on the restart.

JamieRuderman (approved):

PR #559, opened 2021-10-12
[DESK-1219] fix double call state ui

mauricio0000:
### Description
[DESK-1219] fix double call state ui

### Test plan

1. Sign in
2. Go to any path
3.  You should see navigation path working correctly

### Ticket

https://remoteit.atlassian.net/browse/DESK-1219

[DESK-1219]: https://remoteit.atlassian.net/browse/DESK-1219?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/frontend/src/hooks/useNavigation.ts
+++ b/frontend/src/hooks/useNavigation.ts
@@ -39,20 +39,15 @@ export function useNavigation(): INavigationHook {
     remoteUI: isRemoteUI(state),
   }))
   const [shouldUpdate, setShouldUpdate] = useState<boolean>(true)
+  const [navigationPath, setNavigationPath] = useState({})
 
   const match = location.pathname.match(REGEX_FIRST_PATH)
   const menu = match ? match[0] : '/devices'
 
-  // Used for remembering the state of a tab - no longer used
-  const recallPath = (selected: string) => {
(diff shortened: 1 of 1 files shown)


frontend/src/hooks/useNavigation.ts
```diff
@@ -39,20 +39,15 @@ export function useNavigation(): INavigationHook {
...
-  const recallPath = (selected: string) => {
-    const stored = navigation[selected]
-    if (!stored || stored === location.pathname) return selected
-    else return stored
-  }
-
   useEffect(() => {
     if (navigation[menu] !== location.pathname) {
-      ui.set({ navigation: { ...navigation, [menu]: location.pathname } })
+      ui.set({ ...navigationPath, navigation: { ...navigation, [menu]: location.pathname } })
```
javierdac, line 49: When you enter navigationPath is empty and if that variable is changed it will not be update.
This is a question, shall we need to add it on the final array of useEffect?

mauricio0000: navigationPath is an object and it is using updated key to override the reducer

PR #574, opened 2021-11-08
DESK-1242 logs info update

mauricio0000:
### Description

DESK-1242 logs info update

### DIFF
--- a/frontend/src/components/EventList/EventMessage.tsx
+++ b/frontend/src/components/EventList/EventMessage.tsx
@@ -1,5 +1,6 @@
 import React from 'react'
 import { ADD_EVENTS_ACTIONS } from '../../shared/constants'
+import { addDeviceName } from '../../shared/nameHelper'
 
 export const EventType = {
   login_state: 'AUTH_LOGIN',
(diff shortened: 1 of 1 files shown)


frontend/src/components/EventList/EventMessage.tsx
```diff
@@ -43,7 +43,7 @@ export function EventMessage({
     case EventType.device_state:
       message = (
         <>
-          <b> {item.target && serviceName !== device?.name ? serviceName : device?.name} </b>
+          <b> {item.target && serviceName !== device?.name ? serviceName : device?.name + ' [ ' + device?.id + ' ] ' }  </b>
```
JamieRuderman, line 46:
```suggestion
import { addDeviceName } from 'nameHelper.ts' // need real path

let name = device?.id
if (device?.name) name = device.name
if (serviceName) name = addDeviceName(name, serviceName)

...

          <b> {item.target && name}  </b>
```

What about something like this? We want to show device name if available and service name if available. and only if those don't exist then show device id.

JamieRuderman (changes requested):

JamieRuderman (approved):

PR #579, opened 2021-11-10
[DESK-0000] fix before initialization unix URL

mauricio0000:
### Description

https://pretagteam.com/question/cannot-access-authreducer-before-initialization

[](https://pretagteam.com/question/cannot-access-authreducer-before-initialization)

https://github.com/remoteit/desktop/blob/df300c19f9291a47992dfb082f8932e389cc3c97/frontend/src/services/remote.it.ts
it is getting an error with node14 on Ubuntu


JamieRuderman (approved):
odd bug!

PR #592, opened 2021-12-08
DESK-1279 Web updates

JamieRuderman:
### Description

- Adding legacy menu links
- Fixing auto launching in portal mode
- Fix clearing connection settings in portal mode and detecting default connection state
- JSON encoding all local storage so the var types don't change on storage
- Handle routing of all connection update calls to the backend or frontend only depending on state
- DESK-1279 include this device in device list query
- Fixing the filter state and allowing to multi select platforms
- safeguarding against invalid service sort states
- Fix display of 'other' connection info


mauricio0000 (approved):
@JamieRuderman looks good.

PR #612, opened 2022-01-26
DESK-1337 organization setup info

JamieRuderman:
### Description

- Updated electron updater
- Auto updater troubleshooting
- Fix force reload of electron window on crash
- Show share device icon button based on permissions
- Linux setup page add curl/wget and org name
- Fix connection loading
- Update device options menu to be included in service options
- Clear connections when removing device
- Adding role to org select menu
- Allow one line registration to an org you are an admin of
- connectd update to 4.14
- remoteit-components update
- restore enabled connections on startup
- Fix connection data updating
- Fix secondary sort when filtering
- Fix org role select display of unlicensed roles
- Update sign in form
- update one line registration code handling

### Ticket

<https://remoteit.atlassian.net/browse/DESK-1337>


backend/src/ConnectionPool.ts
```diff
@@ -193,7 +198,7 @@ export default class ConnectionPool {
     const json = this.toJSON()
     if (!user.signedIn) return
     this.file?.write(json)
-    d('CONNECTION POOL UPDATED')
+    // Logger.info('CONNECTION POOL UPDATED')
```
javierdac, line 201: remove comment

javierdac (approved):

PR #621, opened 2022-02-11
[Desk-1363] fix 401 errors on bad clock

evanrbowers:
### Description

fix 401 errors on bad clock

### Test plan

Test api calls

### Ticket

<https://remoteit.atlassian.net/browse/Desk-1363>


frontend/src/services/graphQL.ts
```diff
@@ -32,14 +35,57 @@ export async function graphQLRequest(query: String, variables: ILookup<any> = {}
   return await axios.request(request)
 }
 
+export async function graphQLRequestWithErrorHandling(query: String, variables: ILookup<any> = {}, parse : any = undefined) {
```
JamieRuderman, line 38: I don't see what is calling this method? Maybe I missed it?

JamieRuderman, line 38:
```suggestion
export async function graphQLRequestWithErrorHandling(query: String, variables: ILookup<any> = {}, parse?: Boolean  = undefined) {
```

frontend/src/services/graphQL.ts
```diff
@@ -32,14 +35,57 @@ export async function graphQLRequest(query: String, variables: ILookup<any> = {}
...
+
+  if (axios.isAxiosError(error)) {
+    if (error.response?.status === 401 || error.response?.status === 403) {
+      console.log('Incrementing error count: ', errorCount)
+      errorCount = errorCount + 1
+      auth.checkSession({ bypassCache: true })
+      if( errorCount > 1000 ) {
+        auth.signOut()
+      }
+      sleep(1000 * errorCount)
```
JamieRuderman, line 82: Does sleep need an await? Otherwise could just use setTimeout
```suggestion
     await sleep(1000 * errorCount)
```

JamieRuderman (commented):
Just some questions! Looks good thanks!

frontend/src/services/graphQL.ts
evanrbowers: I was going to use it but turned into too much code change for this release.  I think it would be good to start converting the calls over to this in the future to reduce code duplication.

evanrbowers: parse will be a function.

JamieRuderman: No problem. Maybe just leave it commented out with a note? I'd hate to have orphaned coded added if it's not used.

JamieRuderman: I think this should work. I try to avoid `any`
```suggestion
export async function graphQLRequestWithErrorHandling(query: String, variables: ILookup<any> = {}, parse?: () => void = undefined) {
```

JamieRuderman (approved):

JamieRuderman (approved):
Nice - looks good!

JamieRuderman (approved):

PR #640, opened 2022-03-10
[DESK-1351] Remove urls and keys and place in env

evanrbowers:
### Description

Remove urls and keys and place in env

### Ticket

<https://remoteit.atlassian.net/browse/DESK-1351>


evanrbowers:
> @evanrbowers Looks good! Where should we keep a copy of the valid .env?

maybe in a secrets store in aws?

PR #657, opened 2022-04-28
Release fixes

JamieRuderman:
- Fixing admin panel view bugs and updating Advantech logo
- Make sure refresh icon spins for entire refresh
- Lock organization to own account in the account section pages
- If in a devices page return to the root when changing organizations
- Fix display of permission names in attributes
- Fix detection of changed state in target host field
- Fix detection of local registration of own device
- Fix missing VIEW permission on default role
- Add org name button on roles page to return user to org page
- Ignore CLI port response -1


frontend/src/components/Plans.tsx
```diff
@@ -58,7 +58,7 @@ export const Plans: React.FC = () => {
   const { plans, license, purchasing } = useSelector((state: ApplicationState) => ({
     plans: state.plans.plans.filter(p => p.product.id === REMOTEIT_PRODUCT_ID),
     purchasing: state.plans.purchasing,
-    license: selectRemoteitLicense(state),
+    license: selectOwnRemoteitLicense(state),
```
JamieRuderman, line 61: @evanrbowers This fixes your issue I believe.

evanrbowers (approved):

PR #664, opened 2022-05-05
Changing p2p only messaging for desktop

JamieRuderman:
### Description

Changing p2p only messaging for desktop

### DIFF
--- a/frontend/src/buttons/ComboButton/ComboButton.tsx
+++ b/frontend/src/buttons/ComboButton/ComboButton.tsx
@@ -4,8 +4,8 @@ import { makeStyles } from '@material-ui/core'
 import { DisconnectButton } from '../DisconnectButton'
 import { ConnectButton } from '../ConnectButton'
 import { DynamicButton } from '../DynamicButton'
+import { windowOpen, isPortal } from '../../services/Browser'
 import { Notice } from '../../components/Notice'
-import { windowOpen } from '../../services/Browser'
 
 type Props = {
   className?: string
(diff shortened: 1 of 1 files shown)


frontend/src/buttons/ComboButton/ComboButton.tsx
```diff
@@ -24,16 +24,22 @@ export const ComboButton: React.FC<Props> = ({ className, ...props }) => {
...
+              </>
+            )}
           </Notice>
-          <DynamicButton
-            {...props}
-            title="Launch Desktop"
-            color="grayDarkest"
-            onClick={() => windowOpen(`${PROTOCOL}connect/${props.service?.id || props.connection?.id}`)}
-          />
+          {isPortal() && (
```
bstrech, line 36: Does this mean that the button text for Desktop doesn't change? Does it just look like the device is offline?

bstrech (commented):
Wondering if the peer to peer connection fails while it is already in progress, it will just re-connect. In my case it was just about 1 minute, but I was actually using it. My pgAdmin tool told me my connection was lost, so I went to the desktop and saw the message, then it cleared up without me doing anything.

frontend/src/buttons/ComboButton/ComboButton.tsx
bstrech: Answer is in the slack thread here. https://remoteit.slack.com/archives/C019MLSJYS1/p1651765291970749

bstrech (approved):

PR #765, opened 2022-11-01
Remove session id from connect query and merge with publicId

JamieRuderman:
### Description

Remove session id from connect query and merge with publicId

### DIFF
--- a/frontend/src/models/connections.ts
+++ b/frontend/src/models/connections.ts
@@ -287,13 +287,12 @@ export default createModel<RootModel>()({
         console.log('PROXY CONNECTED', data)
         setConnection({
           ...proxyConnection,
-          publicId: data.id,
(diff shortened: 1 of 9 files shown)


frontend/src/models/connections.ts
```diff
@@ -287,13 +287,12 @@ export default createModel<RootModel>()({
...
         setConnection({
           ...proxyConnection,
-          publicId: data.id,
           connecting: false,
           connected: true,
           error: undefined,
           isP2P: false,
           startTime: data.created,
-          sessionId: data.session?.id,
+          sessionId: data.id.toLowerCase(),
```
penngrove, line 295: should not be needed, but fine

JamieRuderman: Fixed!

PR #875, opened 2023-11-29
Initial github workflow for mobile builds

robomon1:
### Test plan

1. Sign in
2. Go to...
3. Click...
4. You should see X, Y, Z...

### Ticket

<https://remoteit.atlassian.net/browse/TICKET_NUMBER_HERE>

### DIFF
--- a/.github/workflows/mobile-build.yml
+++ b/.github/workflows/mobile-build.yml
@@ -7,8 +7,8 @@ defaults:
     shell: bash
 
 env:
-  CI: ''
-  SKIP_SIGNING: true
+  # CI: ''
+  # SKIP_SIGNING: true
 
 jobs:
   release:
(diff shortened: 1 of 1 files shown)


.github/workflows/mobile-build.yml
```diff
@@ -0,0 +1,93 @@
...
+
+on: workflow_dispatch
+
+defaults:
+  run:
+    shell: bash
+
+env:
+  CI: ''
+  SKIP_SIGNING: true
```
JamieRuderman, line 11:
```suggestion
```

We don't need the SKIP_SIGNING as that only relates to electron packaging. I'm not sure what clearing the CI var does, but I suspect it might not be needed either?

JamieRuderman (approved):
Thanks @robomon1 !  Up to you if you wanna remove the ENV stuff.

.github/workflows/mobile-build.yml
robomon1: Commented out for now.

PR #1082, opened 2026-04-30
feature: add support for getting script run logs, add support for bulk updates

evanrbowers:
feature: add support for getting script run logs, add support for bulk status updates over ws

### DIFF
--- a/frontend/src/models/jobs.ts
+++ b/frontend/src/models/jobs.ts
@@ -133,11 +133,15 @@ export default createModel<RootModel>()({
     },
     async downloadLogs({ jobId, jobDeviceId }: { jobId: string; jobDeviceId: string }) {
       const result = await getJobLogs(jobId)
-      if (!result) {
+      if (result.kind === 'error') {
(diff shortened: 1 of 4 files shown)


evanrbowers:
@codex review

JamieRuderman:
One thing worth confirming before merge: in `triggerBrowserDownload` (`frontend/src/services/jobLogs.ts:62`), the `a.download = filename` attribute is [silently ignored on cross-origin URLs](https://html.spec.whatwg.org/multipage/links.html#downloading-resources) per the HTML spec. Since the presigned S3 URL is a different origin, the browser will only force a download (vs. navigating to the file) if the presigned URL itself encodes `response-content-disposition=attachment; filename=...`, or the S3 object's stored `Content-Disposition` is `attachment`.

If the backend isn't already setting that, `.zip` / `.tar.gz` may navigate instead of download on some browsers (Safari especially). Worth a quick check.

JamieRuderman (approved):

PR #1091, opened 2026-05-09
feat(runs): paginate /runs page and link from script page

JamieRuderman:
## Summary
- Paginate the `/runs` page with a Load More button (50 per page); the GraphQL `jobs` query now sends `from`/`size` and reads `total`. Filtered views via `/runs/:fileID` paginate the same way.
- Add a "View all runs" text link at the bottom of the Runs section on the script page → `/runs/:fileID`.
- Show absolute date+time alongside the existing relative duration in the runs Time column.
- Refactor the existing device `LoadMore` into a presentational component shared by `DeviceLoadMore` and `JobLoadMore`.
- Fix a latent crash in `selectJobs` when the per-account cache hasn't been initialized.

## Test plan
- [ ] `/runs` loads, shows total count, Load More fetches the next 50 and merges them
- [ ] `/runs/<fileID>` filters to that script's runs and paginates independently
- [ ] Refresh button on `/runs/<fileID>` preserves the file filter
- [ ] Each row shows month-day clock-time + "X ago"
- [ ] Script page → "View all runs" navigates to `/runs/<fileID>`
- [ ] Device list and Service list Load More still work

### DIFF
--- a/frontend/src/buttons/RefreshButton/RefreshButton.tsx
+++ b/frontend/src/buttons/RefreshButton/RefreshButton.tsx
@@ -38,6 +38,7 @@ export const RefreshButton: React.FC<ButtonProps> = props => {
   const adminUsersPage = useRouteMatch('/admin/users')
   const adminPartnersPage = useRouteMatch('/admin/partners')
   const scriptingPage = useRouteMatch(['/script', '/scripts', '/runs'])
+  const runsPage = useRouteMatch<{ fileID?: string }>('/runs/:fileID?')
   const scriptPage = useRouteMatch('/script')
 
   let title = 'Refresh application'
@@ -57,7 +58,8 @@ export const RefreshButton: React.FC<ButtonProps> = props => {
   } else if (scriptingPage) {
     title = 'Refresh scripts'
     methods.push(dispatch.files.fetch)
-    methods.push(dispatch.jobs.fetch)
+    const runsFileID = runsPage?.params.fileID
+    methods.push(async () => await dispatch.jobs.fetch({ fileID: runsFileID }))
 
     // script pages
   } else if (scriptPage) {
--- a/frontend/src/components/DeviceList.tsx
+++ b/frontend/src/components/DeviceList.tsx
@@ -11,7 +11,7 @@ import { DeviceListItem } from './DeviceListItem'
 import { Attribute } from './Attributes'
 import { isOffline } from '../models/devices'
 import { GuideBubble } from './GuideBubble'
-import { LoadMore } from './LoadMore'
+import { DeviceLoadMore } from './LoadMore'
 import { GridList } from './GridList'
 
 const GUIDE_START_DATE = new Date('2022-09-20')
@@ -141,7 +141,7 @@ export const DeviceList: React.FC<DeviceListProps> = ({
           />
         )
       })}
-      <LoadMore />
+      <DeviceLoadMore />
     </GridList>
   )
 }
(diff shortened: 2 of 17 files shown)


JamieRuderman:
@codex review

JamieRuderman:
## Overview
Adds pagination to `/runs` (and `/runs/:fileID`) using GraphQL `from`/`size` + `total`, refactors the existing `LoadMore` into a presentational component shared by `DeviceLoadMore` and `JobLoadMore`, adds a "View all runs" link from the script page, and shows absolute date+time in the Time column. Also clears stale resource IDs from `/runs`/`/script`/`/file` URLs on org switch.

## 🔴 Bug: Load More inserts page-2 in wrong order at wrong position

In `frontend/src/models/jobs.ts`, the `fetch` effect routes paginated loads through `setJobs`:

```ts
const merge = from > 0 || !!fileID
if (merge) dispatch.jobs.setJobs({ accountId, jobs })
else dispatch.jobs.setAccount({ accountId, jobs })
```

But `setJobs` **unshifts** each new item:

```ts
jobs.forEach(job => {
  const index = updated.findIndex(j => j.id === job.id)
  if (index === -1) updated.unshift(job)
  else updated[index] = job
})
```

For "load older runs" pagination this is doubly wrong:
1. The new page is prepended *above* the existing first page — older items appear on top of newer ones.
2. Within the new page, iterating `[A, B, C]` and unshifting each produces `[C, B, A, ...]` — the page itself ends up reversed.

**Fix:** for pagination merges (`from > 0`), append (`concat`/`push`) instead of unshift, ideally via a dedicated reducer (e.g. `appendJobs`) so `setJobs` retains its current "newest-first upsert" semantics for `fetchSingle`/`fetchByFileIds`.

Worth a manual test: load `/runs`, click Load More, verify the next 50 appear *below* the first 50 in chronological order.

## 🟡 Filtered fetch (`/runs/:fileID`, `from=0`) also goes through merge

Same `merge = from > 0 || !!fileID` branch — initial filter fetch upserts into the global cache. Combined with the unshift above, any filtered items not already cached get prepended (and reversed), which can perturb `/runs` ordering when you navigate back. Worth confirming after fixing the issue above.

## 🟡 Organizational inconsistency for LoadMore variants

- `frontend/src/components/LoadMore/DeviceLoadMore.tsx` lives in `LoadMore/` and is re-exported from `LoadMore/index.ts`.
- `frontend/src/components/JobLoadMore.tsx` lives at the top level and isn't part of the barrel.

Move `JobLoadMore` into `components/LoadMore/` and export it from `index.ts` for consistency.

## 🟢 Smaller notes

- `JobAttributes.tsx`: `Date.now() - date.getTime() < 24h` is computed during render, so a row crossing the 24h boundary keeps the relative line until something else triggers a re-render. Acceptable, but worth a comment if intentional.
- `RefreshButton.tsx`: refresh on `/runs/:fileID` correctly preserves the filter and resets `from` to 0. On `/script` and `/scripts`, `runsFileID` is undefined, so refresh resets jobs pagination on those pages — probably fine but worth confirming it's intentional.
- `selectors/scripting.ts`: the `(jobs[accountId] || [])` guard is a real fix — previously `.filter` on `undefined` would throw when the per-account cache wasn't initialized. Good catch.
- `OrganizationSelect.tsx`: `CLEAR_ID_BASE` lookup uses `menu` as the key — ensure `menu` is the route prefix (`/runs`, `/script`, `/file`) and not a deeper path; otherwise the lookup silently misses.
- The PR description's test-plan checkboxes are all unchecked — please tick the ones you've verified, especially "Load More fetches the next 50 and merges them" given the ordering concern above.

## Security / performance
- No new external input or auth surface; GraphQL variables are typed.
- `structuredClone` of the full job list on every `setJobs` is O(n) per page load — fine at 50 items per page, would matter if `size` grew.

## Recommendation
Block on the load-more ordering bug; the rest are nits. Once `setJobs` no longer reverses paginated pages (or pagination uses an append path), this is good to merge.

PR #1113, opened 2026-06-17
fix(devices): apply active list filter to newly arrived devices

JamieRuderman:
## Problem

When the device list is filtered to a subset (filter drawer, service-type tab, or search) and a new device arrives — registered, shared, or transferred — it was **always prepended to the top with the green NEW badge**, regardless of whether it belongs in the current view. Filtering is done server-side, so the single-device fetch path (`fetchSingleFull`) bypassed all filters.

## Change

Add a client-side matcher that mirrors the server-side filter, and only surface a newly-arrived device (NEW badge, count increment, prepend) when it actually matches the active filter. Non-matching devices are suppressed and reappear naturally on the next refetch (every filter change already triggers one).

Per the desired behavior: matching devices still prepend to the top with the NEW badge — even if a sort would place them off-screen, showing them at the top is wanted. No special sort handling; the only question is filter membership.

### Details
- **`selectors/devices.ts`** — new pure `deviceMatchesFilters(device, model, userId)` next to `selectIsFiltered`. Covers every filter axis: state (online/offline), owner (me/others), platform, tags (ALL/ANY), the service-type tab (`applicationTypes` → device must have a service of a selected type), and a best-effort device-name match for the search query (gated on `searched`, since the live query input can be ahead of the displayed results). Exports the now-shared `IDeviceState` type.
- **`models/devices.ts`** — `fetchSingleFull` gates the `newDevice` treatment on the matcher; the non-`newDevice` path is unchanged.
- **`models/accounts.ts`** — `setDevice` gains a `suppressAdd` flag that blocks only the add-when-absent branch, so a filtered-out new device isn't added while an already-present device is still updated in place.

## Notes
- Search matching is best-effort (device name only — the server also matches service names), so it can occasionally hide a device search would have shown; it reappears on the next refetch. This tradeoff was chosen intentionally.
- `tsc --noEmit` passes clean. No automated test harness exists in this package; end-to-end verification needs a real device registration/share against each filter (per-filter manual checklist available).

### DIFF
--- a/frontend/src/buttons/RefreshButton/RefreshButton.tsx
+++ b/frontend/src/buttons/RefreshButton/RefreshButton.tsx
@@ -7,12 +7,10 @@ import { Dispatch, State } from '../../store'
 import { VALID_JOB_ID_LENGTH } from '../../constants'
 import { useParams, useRouteMatch } from 'react-router-dom'
 import { selectDeviceModelAttributes, selectDevice } from '../../selectors/devices'
-import { selectLimit } from '../../selectors/organizations'
 import { useDispatch, useSelector } from 'react-redux'
 import { IconButton, ButtonProps } from '../IconButton'
 import { GuideBubble } from '../../components/GuideBubble'
 import { Typography } from '@mui/material'
-import { limitDays } from '../../models/plans'
 
 export const RefreshButton: React.FC<ButtonProps> = props => {
   const dispatch = useDispatch<Dispatch>()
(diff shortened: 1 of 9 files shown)


JamieRuderman:
@codex review

JamieRuderman:
Thanks @codex — both suggestions were valid and are now fixed in 6a49cc35:

**P2 · Owner filter relative to account** — Confirmed against the adaptor's own rule (`shared: accountId !== owner.id` in `graphQLDeviceAdaptor`). The matcher now compares `device.owner?.id` to the **active accountId** instead of the logged-in user id, so Owner=Me/Others is correct for organization accounts (it was only coincidentally right for personal accounts where `accountId === userId`).

**P2 · Use the submitted search term** — Right, gating on `searched` didn't cover editing the box after a prior submit. Added an `appliedName` field to the device model, set in `fetchList` to the `query` the list was actually fetched with, and the matcher now tests new devices against `appliedName` rather than the live `query` input. The draft term no longer affects matching until it's re-submitted.

The device-name-only nature of the search match (server also matches service names) remains an intentional best-effort tradeoff. `tsc --noEmit` is clean.

PR #1122, opened 2026-06-24
Add multi-use vs one-time registration code controls on add device.

evanrbowers:
Update existing registration codes in place when options change, keep the command visible during updates, and add a chip toggle for one-time use on the platform add page.

This also retires the product **lock** concept in favor of freely editable products, and adds product tagging plus a "create product from registration" flow.

### Add Device / Platform page

- Added a clickable **MULTI USE / ONE-TIME USE** chip toggle below the tags editor. Defaults to multi-use; toggling controls whether the generated registration code is reusable or expires after a single registration.
- Added a **MAKE PRODUCT** button that creates a reusable device product from the current platform, tags, and selected services, then navigates to the new product's detail page. Disabled while creating or when no services are selected.
- The one-time-use setting is threaded through to the displayed registration code/command and platform overrides (e.g. Rent-a-Node).

### Registration code behavior

- Registration codes are now updated **in place** when tags/services change instead of being cleared and regenerated. The session code is reused and only reset when the account or platform changes, so the command stays visible during edits.
- Redux state is only written when the code/command actually changes, avoiding unnecessary re-renders.
- `createRegistration` now returns `{ registrationCode, registrationCommand }`; callers updated accordingly.

### Products — lock concept removed

- Removed the product **Lock** setting and the NEW/LOCKED status entirely. Products are now always editable.
- Registration command/code on the product settings page now shows unconditionally (no longer gated on the product being locked).
- Removed the product **Status** chip/column from product lists, detail views, and the product title.
- "Add Service" is always available; removed the "Product is locked" blocking screen on the service add page.
- "Remove Service" is no longer hidden on locked products.
- Added an inline **product tag editor** (`ProductTagEditor`) on the product page so tags can be added/removed directly.
- Updated empty-state copy: "Add at least one service before locking the product" → "Add at least one service to this product."

### API / models

- Added `oneTimeUse`, `code`, and `unique` parameters to the registration GraphQL query.
- Added `createDeviceProductFromRegistration` and `updateDeviceProduct` mutations; `IDeviceProduct` now carries `tags` and `source`.
- Replaced `updateSettings` (lock) with `updateProduct` (name/platform/tags) and `createFromRegistration`.

### DIFF
--- a/frontend/src/assets/ConveyorBeltBoxes.tsx
+++ b/frontend/src/assets/ConveyorBeltBoxes.tsx
@@ -1,0 +1,23 @@
+import React from 'react'
+import { useSelector } from 'react-redux'
+import { State } from '../store'
+
+// FontAwesome Pro 7 "conveyor-belt-boxes" (duotone solid), recolored to match
+// the platform icons (e.g. Synology) with a two-tone gray fill.
+export const ConveyorBeltBoxes: React.FC<React.SVGProps<SVGSVGElement>> = props => {
+  const darkMode = useSelector((state: State) => state.ui.themeDark)
+  const light = darkMode ? '#444' : '#bbb'
+  const dark = '#808080'
+  return (
+    <svg viewBox="0 0 640 640" version="1.1" xmlns="http://www.w3.org/2000/svg" {...props}>
+      <path
+        fill={light}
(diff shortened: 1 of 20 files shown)


evanrbowers:
@codex review

evanrbowers:
@codex review

evanrbowers:
@codex review

JamieRuderman (approved):

PR #1130, opened 2026-07-13
fix(auth): verify session before completing two-factor account recovery

JamieRuderman:
## Summary

Fixes the two-factor account recovery loop reported by a locked-out user: completing recovery (email code + recovery code) appeared to succeed but bounced them back to the sign-in screen, where the account still demanded SMS codes that never arrive.

- **`verifyRecoveryCode` now confirms Cognito actually issued a session** (`signInUserSession`) before declaring success. Previously any custom-auth flow that ended without tokens was treated as success, triggering a redirect into an unauthenticated app — the reported loop.
- **A rejected email verification code now stops the flow with a clear error.** Previously the recovery code was blindly submitted as the next challenge answer, so a bad email code silently consumed the recovery code as another email-code attempt and burned Cognito's limited retries.
- **Successful recovery now signs in through `onSignInSuccess`** (same path as the SMS/TOTP flows) instead of a raw `window.location.href = '/#devices'` reload — which was fragile and the wrong format for `HashRouter` (`#devices` vs `#/devices`).

## Not covered here (backend)

- If the recovery lambda ends the flow without issuing tokens or without clearing the SMS MFA preference, users will now see an explicit error instead of a silent bounce — but the lambda side still needs verification.
- SMS delivery failures (SNS logs / spend limit / Canadian origination for +1-705 numbers) are outside this repo.

## Test plan

- `npm run typecheck` passes
- Recovery flow: wrong email code → error shown, recovery code not consumed
- Recovery flow: valid codes → authenticated app renders (no reload), lands on /devices
- Recovery flow: flow ends without session → explicit error instead of redirect to login

### DIFF
--- a/frontend/src/cognito/auth.ts
+++ b/frontend/src/cognito/auth.ts
@@ -343,10 +343,12 @@ export class AuthService {
       // Update the local copy of the Cognito user
       this.cognitoUser = newUser
 
-      // If only a the email verification code was provided,
-      // we assume that the user does not have two-factor setup
-      // and thus does not need the secondary challenge.
-      // if (recoveryCode) {
+      // A rejected email code leaves the EMAIL_CODE challenge pending. Stop here
+      // so the recovery code below isn't consumed as another email code attempt.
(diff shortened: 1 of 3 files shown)


JamieRuderman:
@codex review

PR #1146, opened 2026-07-21
Add persistent BANNER notice type for maintenance and outage info

JamieRuderman:
## Summary
- Add a `BANNER` notice type that renders as a **non-dismissible** bar pinned to the top of the app, intended for system maintenance and outage information.
- Query the previously-unused `preview` (short-form HTML) and `until` fields from the `notices` API and use them for banner copy and expiry.
- Exclude banners from the announcements list, the unread badge, and the full-screen announcement dialog.

## Why
Maintenance and outage notices need to stay visible for the duration of the event. Dismissal is deliberately not offered — a user who dismisses and forgets is the one who files the support ticket the banner was meant to prevent.

Because there is no client-side escape hatch, the banner must reliably remove itself. That is handled server-side by `Notice.visible()`, which already filters on `enabled`, `stage`, `language` and the `from`/`until` window — so ops can expire or kill a banner without a client release.

## Implementation notes
- **`preview` instead of `body`.** The schema already exposes a dedicated short-form "Preview HTML" field, which suits a single-line bar far better than the full `body` HTML. Falls back to `title` when `preview` is empty.
- **`until` is re-checked client-side on every render**, not in the memoized selector. The server already filters expired notices, but `announcements` is in the redux-persist whitelist, so an offline or failed fetch would otherwise leave an expired banner pinned with no way to dismiss it.
- **Reuses the existing `Notice` component** (`severity="warning"`, `fullWidth`, `solid`). Omitting its optional `onClose` is what makes the bar non-dismissible.
- **Multiple active banners stack.** Showing only the newest could hide a second concurrent incident.
- Placed next to `ViewAsBanner` in `App.tsx`, matching the existing top-banner pattern.

## ⚠️ Requires an API change before it does anything
`NoticeType` in `graphql-api` does not yet include `BANNER`. Until it does, no notice can carry the type and this code renders nothing (it is inert, not broken).

The enum is stored as an `int` (`GENERIC=0` … `COMMUNICATION=4`), so appending `BANNER=5` is purely additive — no DB migration needed, the column is already `int`, and existing rows are not renumbered.

`preview` and `until` are already exposed as GraphQL fields, so querying them is safe today.

## Validation
- `npm run typecheck`
- `npm run build`
- Not yet exercised end-to-end in the running app, since no notice can return `type: BANNER` until the API enum lands.


JamieRuderman:
## CodeQL: incomplete multi-character sanitization — fixed in `4c65b611`

The alert on `adminNoticeAttributes.tsx` was against my `plainText` helper:

```ts
const plainText = (html?: string) => html?.replace(/<[^>]*>/g, '').trim()
```

**Not currently exploitable** — its only call site feeds a React text child (the list's Body column), which React escapes. But CodeQL is right that it's a booby trap: it reads like a sanitizer and isn't, so any future use near `dangerouslySetInnerHTML` would be a real vulnerability.

Confirmed the bypass in Chromium rather than assuming:

| input | old regex | fixed |
|---|---|---|
| `<script src=x` | `<script src=x` ← **passed through intact** | `` |
| `<<script>script>alert(1)` | `script>alert(1)` | `<` |
| `<script>alert(1)</script>` | `alert(1)` ← JS source leaked into the cell | `` |
| `<style>body{}</style>hi` | `body{}hi` ← CSS leaked | `hi` |
| `<p>Full <b>HTML</b> body copy.</p>` | `Full HTML body copy.` | `Full HTML body copy.` |
| `plain text` | `plain text` | `plain text` |

An unterminated tag never matches `<[^>]*>`, so it survived untouched — exactly the class of hole the rule exists for.

Replaced with DOM parsing, which is correct by construction:

```ts
const doc = new DOMParser().parseFromString(html, 'text/html')
doc.body.querySelectorAll('script, style').forEach(el => el.remove())
return doc.body.textContent?.trim() || ''
```

`parseFromString` builds an **inert** document — scripts don't run, `onerror` never fires. The `script, style` removal is a second thing the regex got wrong: `textContent` otherwise includes their *source*, so a notice body could spill JS or CSS into a list cell.

Normal input is unchanged; only the adversarial cases differ. Comment on the function now states explicitly that it is display-only and must never be used to produce HTML.

`tsc --noEmit` clean, build clean.

---

### Related, not fixed: the banner/card HTML sink

Worth flagging since it came up in the graphql-api#171 review too. `AnnouncementBanner` renders `body` via `dangerouslySetInnerHTML` with no sanitization, matching the pre-existing `AnnouncementCard`. That's an accepted property of the notice system — HTML-by-design, authored by super-admins only (`UserRole.ADMIN` → `yoics.admin`, internal staff, **not** org admins) — so it's not a privilege-escalation path and not a regression from this PR.

It does mean a compromised staff account is a stored-XSS vector against the whole user base, and the banner is more prominent than the card. If you want defense-in-depth, the fix is central sanitization at parse time in the announcements model rather than at each render site. Happy to do it as a follow-up; deliberately not folded into this PR since it would change existing card behaviour.

PR #1147, opened 2026-07-22
feat(i18n): multi-language support (en/ja/de/es) foundation

JamieRuderman:
## Summary

Adds a working multi-language (i18n) foundation across all three platforms (web portal, Electron desktop, Capacitor mobile) plus the Electron main process. Supports **English** (source/fallback) + **Japanese, German, Spanish**, with a language picker and OS-based auto-detection.

Because untranslated strings fall back to English, the app is shippable at every step — this PR lands the full infrastructure and a proven extraction template; the remaining bulk string extraction can follow incrementally.

## What's included

- **Core infra** — app-wide i18next instance (`frontend/src/i18n/`) with `app`/`notices`/`cognito` namespaces. English bundled eagerly; ja/de/es lazy-loaded and code-split per locale. The existing Cognito auth module is folded into the shared instance.
- **Language preference** — `language` state in `models/ui.ts` mirroring the existing `themeMode` pattern (localStorage, persists across logout). Default `system` follows the OS/browser/device language via `navigator.language`, re-resolved each launch.
- **Auto-detection** — `system` (the default) matches the OS language on all three platforms; region tags degrade gracefully (`ja-JP`→`ja`, unknown→`en`). Standard "follow OS + in-app override" convention.
- **Picker UI** — selector in `OptionsPage.tsx` with native endonyms (日本語, Deutsch, Español).
- **Electron main** — tray, menus, and native dialogs localized (`electron/src/i18n.ts` + `electron/src/locales/*.json`); language synced from the renderer through the existing preferences channel (no new IPC).
- **Dates/durations** — `humanizeDuration()` + `getLocale()` in `dateHelper.ts`; luxon locale follows the app language.
- **Tooling** — `npm run i18n:extract` (i18next-parser) and `npm run i18n:check` (CI key-parity / empty-English gate).

Cognito auth and OptionsPage are fully translated in all four languages (machine-seeded, pending professional review before production promotion).

## Verified

- Type-check clean (frontend + electron), production build passes with per-locale code-splitting.
- Runtime: live language switching confirmed in a browser harness — interpolation (`{{device}}`, `{{version}}`) renders correctly in Japanese with adapted word order; unseeded keys fall back to English (`returnEmptyString: false`).
- `i18n:check` passes across all 4 locales / 3 namespaces.

## Remaining (follow-up work)

- Bulk extraction of ~570 more `.tsx` files (order: shell/nav → core surfaces → ~41 redux-model notice messages with plural fixes → long tail). Each batch: extract → parser → Sonnet-seed → parity gate.
- Pseudo-locale + key-parity CI job; full manual matrix (Electron tray, iOS simulator).

## Decisions worth noting

- **Strict TS key-typing rejected** — it forces `t()` return types into a union that breaks existing string usages and dynamic keys; the CI parity check covers correctness instead.
- **Parser set to never auto-prune** — some cognito keys are built dynamically and would otherwise be silently deleted.


JamieRuderman:
@codex review

JamieRuderman:

PR #1152, opened 2026-07-24
feat(i18n): extract shell navigation strings (batch 1)

JamieRuderman:
## Summary

First string-extraction batch (Phase 4): the **shell navigation chrome** — the sidebar nav, the avatar/account menu, and the mobile bottom nav. These labels appear on every screen, so they're the highest-visibility surface to internationalize first and the best proof of the pattern at scale.

**Targets `feature/i18n`**, not `main` — this stacks on the i18n foundation and is meant to be reviewed on its own.

## Changes

- **`SidebarNav.tsx`, `AvatarMenu.tsx`, `BottomMenu.tsx`** — all hardcoded labels, tooltips, and confirmation-dialog copy routed through `t()` under a new `nav.*` key group (31 keys). Shared labels (Connections/Devices/Networks) reuse the same keys across sidebar and bottom nav; the "Are you sure?" confirm titles reuse the existing `common.areYouSure`.
- **Interpolation, not concatenation** — the two count tooltips use placeholders (`{{connections}} Connections - {{active}} Connected`), and the old `(testUI ? '' : 'Enable ') + 'Test UI'` string-building was split into two full-sentence keys (`nav.testUI` / `nav.enableTestUI`).
- **Translations** — `nav.*` seeded in ja/de/es (machine-translated via Sonnet, flagged for professional review before production). Placeholder parity verified: `{{connections}}`/`{{active}}` preserved in all three locales.

## Tooling fix (separate commit)

Found and fixed a bug in `i18next-parser.config.js`: the installed parser passes the inline `t()` default as a **plain string** in the callback's 4th arg, but the config read it as an object (`value?.usageContext?.defaultValue`), so English values always came out empty. Now English catalogs populate correctly from the inline defaults. (The earlier `options.*` values had survived only because the parser preserves pre-existing non-empty values.)

## Verification

- `i18n:check` passes — 4 locales × 3 namespaces, no missing keys, no empty English.
- `tsc --noEmit` clean; production build passes.
- Placeholder parity across en/ja/de/es confirmed programmatically (zero mismatches).

Untranslated strings elsewhere still fall back to English, so nothing regresses.


JamieRuderman:
@codex review

JamieRuderman:

PR #1153, opened 2026-07-24
feat(i18n): extract redux notice messages (batch 2)

JamieRuderman:
## Summary

Batch 2 of Phase 4 extraction: the **redux model notice messages** — the success/error/notice toasts dispatched from `dispatch.ui.set({ successMessage / errorMessage / noticeMessage })` across 16 model files. 60 messages → 65 keys in the `notices` namespace, fully translated to ja/de/es.

Since these run in non-React model code, they call the shared `i18n` instance directly: `i18n.t('notices:key', { ...vars, defaultValue })`.

**Stacked on `i18n/extract-shell` (#1152)** because it depends on that branch's parser fix. Once #1152 merges to `feature/i18n`, I'll retarget this PR — the diff here is notices-only.

## Highlights — the pitfalls the plan called out, handled

- **Pluralization** — replaced hand-rolled ternaries (`device${n>1?'s were':' was'}`) with i18next `count` plurals (`_one`/`_other`). This also *fixes an English bug*: the old code rendered "1 devices were…"; it now correctly says "1 device was…". Verified at runtime across locales (German inflects the verb too: "wurde" vs "wurden"; Japanese uses its single form).
- **Interpolation, never concatenation** — every `${name}`/`${email}`/`${error}` became a `{{placeholder}}`; string-built messages (`'Update phone error: ' + error.message`) collapsed into one keyed call.
- **Word order** — the possessive `${who} ${plan} subscription updated` (where `who` was `'Your'` or `` `${email}'s` ``) was split into two keys (`subscriptionUpdatedYours` / `subscriptionUpdatedOther`) so each language renders possession naturally rather than forcing English `'s`.
- **Conditional nouns** — `that ${isService ? 'service' : 'device'}` split into two keys.
- **Not translated** — API/dynamic values (device names, emails, `error.message`) stay as runtime interpolation values.

## Tooling fix (separate commit)

Made `i18n:check` **plural-aware**. Plural keys expand per-locale via CLDR (Japanese: `_other` only; English/German: `_one`+`_other`; Spanish also `_many`), so strict key-equality would wrongly flag Japanese as "missing" `_one`. The check now compares plural *bases* against each locale's own `Intl.PluralRules` categories.

## Verification

- `i18n:check` passes (plural-aware) — 4 locales × 3 namespaces, no missing/dead keys, no empty English.
- `tsc` clean; production build passes.
- Placeholder parity verified programmatically across en/ja/de/es (zero mismatches).
- Runtime plural resolution spot-checked: count=1 vs 5 renders the correct grammatical form in every language.

### DIFF
--- a/frontend/scripts/i18n-check.mjs
+++ b/frontend/scripts/i18n-check.mjs
@@ -36,20 +36,58 @@ const report = msg => {
   console.error(`  ✗ ${msg}`)
 }
 
+// Plural keys (key_one, key_other, …) expand per-locale: Japanese has only
+// `other`, English/German/Spanish have `one`+`other`. So the set of concrete
+// keys legitimately differs between locales — compare by plural BASE plus each
+// locale's own CLDR categories, not by raw key equality.
+const PLURAL_SUFFIX = /^(.*)_(zero|one|two|few|many|other)$/
+const splitPlural = key => {
+  const m = key.match(PLURAL_SUFFIX)
+  return m ? { base: m[1], cat: m[2] } : { base: key, cat: null }
+}
+const categoriesFor = locale => new Set(new Intl.PluralRules(locale, { type: 'cardinal' }).resolvedOptions().pluralCategories)
+
 for (const ns of namespaces) {
   const source = flatten(load(SOURCE, ns))
-  const sourceKeys = new Set(Object.keys(source))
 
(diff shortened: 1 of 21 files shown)


JamieRuderman:
@codex review

JamieRuderman:

PR #1155, opened 2026-07-24
feat(i18n): extract device list chrome (batch 4)

JamieRuderman:
## Summary

Batch 4 of Phase 4 — the first slice of the **device/connection surfaces**: the device-list **chrome**. 20 keys (`deviceList.*`) across `DevicesHeader`, `DeviceListEmpty`, and `DevicesActionBar`, translated to ja/de/es.

Covers the empty/no-results states, the AWS demo-device onboarding copy, the Cachengo-brand welcome, and the multi-select action bar (tag/script controls, selection labels).

**Stacked on `i18n/extract-settings` (#1154).**

## Notes / scoping

- **Column labels deferred deliberately.** The device-list column headers come from `Attributes.tsx` — a module-level static registry of ~71 `label:` strings evaluated once at import. Making those translatable and reactive to language changes needs a small architectural change (labels → keys resolved at render time), so it gets its own dedicated batch rather than being rushed in here.
- **Conditional noun split** — `Your {searched ? 'search' : 'filter'} returned no results` became two keys so each language reads naturally.
- **Brand copy** — the Cachengo/Telepath strings render only for that brand build; product names (`Telepath`, `Cachengo`, `Symbiote Node`, `Rent-A-Node`) are kept untranslated, surrounding words localized.
- `DevicesPage` and `DeviceListHeaderCheckbox` had no user-facing strings (pure composition), so they're untouched.

## Verification

- `i18n:check` passes — 4 locales × 3 namespaces, no missing/dead keys, no empty English.
- `tsc` clean; production build passes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1156, opened 2026-07-24
feat(i18n): translatable Attribute column labels (batch 5)

JamieRuderman:
## Summary

Batch 5 — makes the **device/service/connection table column labels translatable**. This is the architectural piece deferred from the device-list chrome batch: the `Attribute` registry in `Attributes.tsx` is a module-level static array (~54 string labels) evaluated once at import, so labels can't use the `t()` hook.

54 `columns.*` keys, translated to ja/de/es. **Stacked on `i18n/extract-devicelist` (#1155).**

## Approach (the option chosen: label → id-keyed getter)

- `Attribute.label` / `Attribute.help` are now **getters** that resolve `i18n.t('columns.<id>')` / `columns.<id>_help` at access (render) time, keyed by the attribute's existing `id`, with the original English string as `defaultValue`. Backing fields `_label` / `_help` hold what the constructor passes.
- **React-node labels pass through unchanged** (e.g. the graph-column headers `<DeviceGraphColumn />`).
- **Fallback is safe**: any id without a catalog entry renders its English default — so the many other `Attribute`-based registries (File/Job/Product/Customer) keep working in English until a later batch, with no code change needed.
- Verified **no id collisions**: those other registries use prefixed ids (`customer*`, `job*`, `product*`, `script*`), so id-keying is unambiguous.

## Catalog mechanics

- Keys are **dynamic** (`columns.${id}`), so the i18next-parser can't discover them — they're **hand-maintained** in the catalogs (same pattern as the cognito dynamic keys). Documented in `i18next-parser.config.js`; `keepRemoved: true` means an extract run leaves them intact (verified).
- The mock placeholder attributes (`categoryA–E`, `statusA–E`) are intentionally *not* keyed — they fall back to English.

## Deferred (follow-up)

Hardcoded English still lives **inside `value` functions** in this file — connection-type strings ("Peer to Peer", "Inactive", "Local Proxy", …), plus "Yes"/"No", "This system", "since refresh". Those render via inline `i18n.t()` (no architecture change) and are a clean separate batch.

## Verification

- `i18n:check` passes; `tsc` clean; production build passes.
- Extraction run confirmed the dynamic `columns.*` keys survive (not pruned).
- Runtime getter check: labels resolve per language (Name→名前/Name/Nombre, External IP→外部IP/Externe IP/IP externa, …), and an unknown id correctly falls back to its English default.


JamieRuderman:
@codex review

JamieRuderman:

PR #1158, opened 2026-07-24
fix(i18n): address Codex review findings

JamieRuderman:
Addresses the Codex review findings across the i18n PR stack. Verified each against source before fixing.

## P1
- **Signed-out screens now follow the OS locale.** `i18n` init resolved `lng` from a hard-coded `'en'` only, and the language was re-resolved solely on sign-in — so translated Cognito sign-in/signup/MFA screens always rendered English. Init now calls `resolveLanguage()` (→ `navigator.language`); the persisted per-account override is session-scoped and still re-applied by `ui.setLanguage` after sign-in.

## P2 — concrete bugs
- **Confirmation dialog buttons** (`Confirm.tsx`) were hard-coded `Ok`/`Cancel` → now `common.ok`/`common.cancel`. Affects every confirm dialog app-wide.
- **Relative dates** (`Duration.tsx`) appended a literal English `" ago"` (`"3 Tage ago"`) and formatted month/year/date branches with `navigator.language`. Now uses a translated `duration.ago` phrase (word-order per locale: `vor {{d}}` / `{{d}}前` / `hace {{d}}`) and `getLocale()`.
- **Empty attribute labels** (`Attributes.tsx`) leaked the raw `columns.<id>` key (e.g. the enterprise-license actions column) under `returnEmptyString:false` → blank labels now pass through untranslated.
- **`networkName`** classified system networks by *name*, so a user network named "Local"/"Recent"/etc. got relabeled. Now gated on the reserved synthetic id (`local`/`recent`/`public`).
- **Gravatar link** ran into preceding text in de/es (`bittegravatar.com`) → added `{' '}`.
- **Sidebar count/tooltip** number formatting now uses `getLocale()`.

## P2 — reactivity (module-level strings didn't re-resolve)
- `RouteSetting.ROUTES` was evaluated at import (while i18n was still `en`) → converted to `getRoutes()` resolved at render (also updated `ServiceAttributesForm`).
- Added `useTranslation()` at the `App` root so render-time translations resolved outside React (Attribute label getters, value functions, date/duration helpers) re-render on language switch and when a lazy locale chunk finishes loading.

## Deliberately deferred
- **`applications.ts` launch-method labels** (`Command`/`Terminal`/…) are set in constructors and memoized by `selectApplication`, so they stay stale on a *live* language switch until the service/connection changes. Fixing this cleanly means a getter refactor in the shared `common/` package; deferred as low-impact (narrow surface, only on live switch) rather than risk an untested shared-package change.
- **Admin table header collisions** (`created` → "Created date", `license`) — left as-is: admin pages are internal-only and don't need translation.

Verification: `tsc` clean, production build clean (locale chunks code-split), i18n parity check passes.

### DIFF
--- a/frontend/src/components/App.tsx
+++ b/frontend/src/components/App.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect } from 'react'
+import { useTranslation } from 'react-i18next'
 import browser from '../services/browser'
 import useSafeArea from '../hooks/useSafeArea'
 import useCapacitor from '../hooks/useCapacitor'
@@ -34,6 +35,11 @@ import { AnnouncementDialog } from './AnnouncementDialog'
 import { AnnouncementBanner } from './AnnouncementBanner'
 
 export const App: React.FC = () => {
+  // Subscribe the whole app to i18next language changes and lazy-locale loads, so
+  // render-time translations resolved outside React (Attribute label getters,
+  // value functions, date/duration helpers) re-render when the language switches
+  // or a non-English catalog chunk finishes loading.
+  useTranslation()
   const { insets } = useSafeArea()
   const location = useLocation()
   const hideSplashScreen = useCapacitor()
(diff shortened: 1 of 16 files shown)


JamieRuderman:
@codex review

JamieRuderman:

PR #1160, opened 2026-07-27
fix(admin): open view-as-user in the web portal from the desktop app

JamieRuderman:
## Problem

The admin **View as** eye icon doesn't work in the desktop app. The desktop app renders data from the local backend, so navigating in-app with the `viewAs` param never produces the target user's view.

## Change

In Electron, the eye icon now launches the web portal in the system browser instead of pushing an in-app route:

- Added `PORTAL_URL` to `frontend/src/constants.ts` (`VITE_PORTAL_URL` → brand `package.homepage` → `https://app.remote.it`), so it can be pointed at a non-production portal.
- `AdminUserDetailPage.handleViewAsUser` opens `${PORTAL_URL}/#/devices?viewAs=<id>,<email>` externally when `browser.isElectron`.

Mobile keeps the in-app `history.push`, and web keeps opening a new tab of itself — both unchanged.

## Notes

- Electron's `windowOpen` goes through `window.open` → `setWindowOpenHandler` in `electron/src/ElectronApp.ts`, which calls `shell.openExternal`, so the default browser handles it.
- The admin must already be signed in at app.remote.it in that browser; the desktop session isn't shared, so they'll hit sign-in first if not.

## Testing

- `tsc --noEmit` passes.
- Not yet click-tested in a running Electron build.


JamieRuderman:
@codex review

PR #1161, opened 2026-07-27
Extract notice list text via DOM parsing instead of a tag regex

JamieRuderman:
Closes CodeQL alert [#31](https://github.com/remoteit/desktop/security/code-scanning/31) — *Incomplete multi-character sanitization*.

Follow-up to #1146, which introduced `plainText`. The alert was raised on that PR but the fix landed on an unrelated branch by mistake, so `main` still carries the original version. This is that fix, cherry-picked onto `main` for clean provenance.

## The problem

```ts
const plainText = (html?: string) => html?.replace(/<[^>]*>/g, '').trim()
```

Used for the Body column of the admin notices list. **Not currently exploitable** — its only call site renders into a React text child, which escapes. But it reads like a sanitizer and isn't, so any future use near `dangerouslySetInnerHTML` would be a live vulnerability. That's the class of bug the CodeQL rule exists for.

Verified in Chromium rather than assumed:

| input | old regex | this PR |
|---|---|---|
| `<script src=x` | `<script src=x` ← **passed through intact** | `` |
| `<<script>script>alert(1)` | `script>alert(1)` | `<` |
| `<script>alert(1)</script>` | `alert(1)` ← JS source in the cell | `` |
| `<style>body{}</style>hi` | `body{}hi` ← CSS in the cell | `hi` |
| `<p>Full <b>HTML</b> body copy.</p>` | `Full HTML body copy.` | `Full HTML body copy.` |
| `plain text` | `plain text` | `plain text` |

An unterminated tag never matches `<[^>]*>`, so it survived untouched.

## The fix

```ts
const doc = new DOMParser().parseFromString(html, 'text/html')
doc.body.querySelectorAll('script, style').forEach(el => el.remove())
return doc.body.textContent?.trim() || ''
```

`parseFromString` builds an **inert** document — scripts don't execute, `onerror` never fires. The `script, style` removal covers a second thing the regex got wrong: `textContent` otherwise includes their *source*, so a notice body could spill JS or CSS into a list cell.

Fast path preserved for the common case (`!html.includes('<')` returns early), so plain-text bodies skip parsing entirely.

The comment on the function now states explicitly that it is display-only and must never be used to produce HTML.

## Scope

Behaviour is unchanged for all normal input; only adversarial/malformed HTML differs.

This does **not** address the separate, larger issue that `AnnouncementCard` and `AnnouncementBanner` render notice `body` through `dangerouslySetInnerHTML` with no sanitization — that's tracked in remoteit/graphql-api#182.

## Validation

- `npm run typecheck` — clean
- `npm run build` — clean
- Bypass cases verified in Chromium (table above)


JamieRuderman:
@codex review

PR #1164, opened 2026-07-30
feat: device detail page in the Admin Devices section

evanrbowers:
Follow-up to #1162: picking a device in the Admin Devices list now opens a **device detail page** at `/admin/devices/:deviceId` instead of jumping straight to the owner's admin user page. The owner is a row on that page — clicking it navigates to the user. So the device is the destination, and the user is one step away rather than the other way around.

**Layout** follows the AdminUsersPage split-panel pattern: resizable list + detail, with the detail taking the full area when there isn't room for two panels, and the same selection persistence (returning to Devices restores the last selected device).

**The detail page shows** what support actually needs: status, platform, owner (clickable), device + hardware ids (copyable), version, license, created / last reported, capabilities, external + internal addresses, geo/ISP, connection quality, the service list with per-service state, and who the device is shared with. Backed by a new `graphQLAdminDevice` query and a `detailCache` in the adminDevices model that merges over the sparse list row, mirroring `adminUsers`.

Also sets `translate = false` on `AdminDeviceAttribute`, matching the convention for internal-only admin registries (as `AdminUserDevicesPanel` already does).

Frontend typechecks clean (0 errors) and the app boots with no console/render errors; the page itself renders behind admin login, so a signed-in pass is worth doing before release.

### DIFF
--- a/frontend/src/components/Header/Header.tsx
+++ b/frontend/src/components/Header/Header.tsx
@@ -44,7 +44,7 @@ export const Header: React.FC<Props> = ({ panels = 1 }) => {
   const menu = location.pathname.match(REGEX_FIRST_PATH)?.[0]
 
   // Admin pages have two-level roots: /admin/users and /admin/partners (without IDs)
-  const adminRootPages = ['/admin/users', '/admin/admins', '/admin/partners', '/admin/enterprise-licenses', '/partner-stats']
+  const adminRootPages = ['/admin/users', '/admin/admins', '/admin/partners', '/admin/enterprise-licenses', '/admin/devices', '/admin/notices', '/partner-stats']
   const isAdminRootPage = adminRootPages.includes(location.pathname)
   const isRootMenu = menu === location.pathname || isAdminRootPage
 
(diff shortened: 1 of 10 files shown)


evanrbowers:
Added the back-navigation fix reported in testing (72f8a112).

Back from the device detail page was landing on the **Users** list. Two pre-existing causes:

1. `ROUTE_PARENTS` had no `/admin/devices/:deviceId` entry, so single-panel UP fell back to the section root `/admin` — which redirects to `/admin/users`.
2. For multi-panel layouts `useNavigationUp` skips the map entirely and goes to the section root unless the section is in `MENU_SECTIONS`. The admin sidebar *is* a menu rather than the section's own list, so `/admin` belongs there. Without it, every admin sub-page walked back to Users regardless of origin — the users pages only looked right because `/admin` happens to redirect to their list.

**Intentional side effect worth a look:** back from a user's account/devices sub-panel now returns to *that user* instead of the users list, matching single-panel behavior and the rest of the app's hierarchy. Easy to scope to devices only if you'd rather not change the users pages.

Also added `/admin/devices` and `/admin/notices` to the header's `adminRootPages`, so those list pages stop showing a back arrow that pointed out of the section.

PR #1169, opened 2026-08-03
refactor(sort): route alphabetical sorts through a shared comparator

JamieRuderman:
Stacked on #1167 — review that one first. GitHub will retarget this to `main` automatically when #1167 merges.

"Alphabetical" meant four different things across the UI. There were ~14 hand-rolled comparators in 12 files, in four flavors:

- bare `localeCompare` (case-sensitive) — `Tags`, `selectedHelper`, `selectors/tags`
- `localeCompare` with `{ sensitivity: 'base' }` — `selectors/contacts`
- `toLowerCase()` with `<`/`>` — `connections`, `search`, `SortServices`, `FilterDrawer`, `OrganizationSelect`, `OrganizationSelectList`, `NetworksPage`
- raw `>`/`<` on unnormalized strings — `SharedUsersLists`

Plus `alphaEmailSort` copy-pasted verbatim into `OrganizationMemberList` and `OrganizationGuestList`.

## Changes

`helpers/utilHelper.ts` gains one shared comparator pair backed by a single `Intl.Collator`:

- `alphaSort(a, b)` — for strings (emails, `nodeName`, nested fields)
- `byName(a, b)` — for the common `{ name }` object case, so most call sites become `.sort(byName)`

Every site above now routes through it, and the two private `nameSort`s plus both `alphaEmailSort` copies are deleted. Net -38/+43 lines across 15 files.

## Behavior changes

This deliberately makes previously-inconsistent sorts agree, so some orderings shift:

- **`numeric: true` everywhere** — `device2` now sorts before `device10` instead of after. This is the main visible change and the reason for the collator.
- **`sensitivity: 'base'`** — case and accents no longer split otherwise-equal names. The `toLowerCase()` sites were already effectively case-insensitive; the bare-`localeCompare` sites (tags, device selection) change.
- **Two-way comparators now return 0 on equal.** `OrganizationSelect`, `OrganizationSelectList`, `FilterDrawer`, and `NetworksPage` used `? 1 : -1`, which never returns 0 and gives an unstable order for equal names. They're now stable.
- **`SortServices` ZTOA** is `byName(b, a)` rather than a separately hand-written reversed comparator.

## Testing

`tsc --noEmit` passes. Verified by grep that no `localeCompare`, `toLowerCase()` comparison, or raw name/email comparator remains in `frontend/src`. Not exercised in the running app — the sorts are visual, so a click through tags, networks, org member lists, and the service sort menu is worth doing before merge.

### DIFF
--- a/frontend/src/buttons/RefreshButton/RefreshButton.tsx
+++ b/frontend/src/buttons/RefreshButton/RefreshButton.tsx
@@ -1,4 +1,5 @@
 import React from 'react'
+import { useTranslation } from 'react-i18next'
 import network from '../../services/Network'
 import cloudController from '../../services/cloudController'
 import cloudSync from '../../services/CloudSync'
@@ -13,6 +14,7 @@ import { GuideBubble } from '../../components/GuideBubble'
 import { Typography } from '@mui/material'
 
 export const RefreshButton: React.FC<ButtonProps> = props => {
+  const { t } = useTranslation()
   const dispatch = useDispatch<Dispatch>()
   const { deviceID, fileID, jobID, userId, partnerId } = useParams<{
     deviceID?: string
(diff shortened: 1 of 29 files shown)


JamieRuderman:
@codex review

